PR #4, opened 2020-10-26
Add JSONError class

c-bata:


optuna_dashboard/app.py
```diff
@@ -50,18 +57,22 @@
...
         try:
             response_body = view(*args, **kwargs)
             return response_body
         except Exception as e:
-            response.status = 500
-            response.content_type = "application/json"
-            stacktrace = "\n".join(traceback.format_tb(e.__traceback__))
-            logger.error(f"Exception: {e}\n{stacktrace}")
-            return json.dumps({"reason": "internal server error"})
+            raise JSONError("internal server error") from e
```
c-bata, line 75: It does not mean anything if disabled debug mode or not returning debug html.

PR #18, opened 2021-01-14
Put select box to change the reload interval time on AppBar.

### LINKED ISSUE #17: Do not reload graphs when zoomed in
Currently, when I zoom into a graph (e.g. intermediate values), the graph is reset as soon as the network fetch in the background completes (every 10s?).

It would be nice if:
* there was a button to stop autoreload
* or the graph interactions would be restored/kept across data reloads

Otherwise, I love it! Thank you so much for the project

c-bata:
To close #17.

I implemented switch button to disable auto-reloading. But I'm still thinking about the UI...

Screenshot 2021-01-15 4 56 59


t-moe:
Wow, that was fast support :).
The location of the button is ok. But you could also put it into the header beneath the "home icon" and keep it visible on all pages (even when not currently needed).

t-moe:
That's wonderful!

PR #34, opened 2021-02-15
Port plot_slice to typescript

2403hwaseer:
# Description
Porting `optuna.visualization.plot_slice` to typescript. Users can now visualize slice plot in dashboard.

# Context
Resolves #28

PR #52, opened 2021-02-28
Remove redundant computations for GraphSlice

c-bata:
Followup of #34 

* Speed up by remove redundant for-loop.
* Allocate graph area even if empty trials.

### DIFF
--- a/optuna_dashboard/static/components/GraphSlice.tsx
+++ b/optuna_dashboard/static/components/GraphSlice.tsx
@@ -22,25 +22,37 @@ const useStyles = makeStyles((theme: Theme) =>
   })
 )
 
+const getParamNames = (trials: Trial[]): string[] => {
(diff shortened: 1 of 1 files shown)


optuna_dashboard/static/components/GraphSlice.tsx
```diff
@@ -103,143 +115,107 @@ const plotSlice = (study: StudyDetail, objectiveId: number, xAxis: string) => {
...
+      gridwidth: 1,
+      tickfont: {
+        color: "#000000",
       },
-      plot_bgcolor: "#E5ecf6",
-      showlegend: false,
+      tickvals: tickvals,
+      ticktext: vocabArr,
     }
-    paramNames.forEach((paramName) => {
```
c-bata, line 174: This loop should be removed.

Issue #25, opened 2021-02-08
[Feature] Hyperparameter importance support


c-bata:
# Feature Request

Add Hyperparameter importance support for direct visualization of which hyperparameters are driving the results.

* [x] Add API endpoint to get hyperparameter importance
* [x] Porting [plot_param_importance()](https://optuna.readthedocs.io/en/stable/reference/visualization/generated/optuna.visualization.plot_param_importances.html#optuna.visualization.plot_param_importances) function to TypeScript.

PR #54, opened 2021-03-10
Add hyperparameter importance chart

chenghuzi:

## Reference Issues/PRs

Fixs #25 

## Add hyperparameter importance chart

There're two ways to add hyperparameter importance chart. One is to implement the evaluator which completely mirrors the one used by optuna via typescript and another is to use the optuna `optuna.importance.get_param_importances` API directly. 

The former makes the logic cleaner but involves extra computation in the browser, and implementing a complete parameter importance function by ts requires an update whenever the optuna API changes.

The latter one is much more flexible and only requires introducing another API call. But the con is it takes longer for the server to get the importance result.

Here we take the second approach and only consider the single objective function case. For future multi-objective case, we can extend the API with query parameters representing params and target selected by the user.

### DIFF
--- a/optuna_dashboard/app.py
+++ b/optuna_dashboard/app.py
@@ -196,7 +196,7 @@ def create_app(storage: BaseStorage) -> Bottle:
     @app.get("/api/studies/<study_id:int>/param_importances")
     @handle_json_api_exception
     def get_param_importances(study_id: int) -> BottleViewReturn:
-        # TODO: add support for selecting params and targets via query parameters.
+        # TODO(chenghuzi): add support for selecting params and targets via query parameters.
         response.content_type = "application/json"
         study_name = storage.get_study_name_from_id(study_id)
         study = Study(study_name=study_name, storage=storage)
--- a/optuna_dashboard/app.py
+++ b/optuna_dashboard/app.py
@@ -99,7 +99,7 @@ def get_distribution_name(param_name: str, study: Study) -> str:
     for trial in study.trials:
         if param_name in trial.distributions:
             return trial.distributions[param_name].__class__.__name__
-    assert False
+    assert False, "Must not reach here."
 
 
 def create_app(storage: BaseStorage) -> Bottle:
--- a/optuna_dashboard/app.py
+++ b/optuna_dashboard/app.py
@@ -217,11 +217,11 @@ def create_app(storage: BaseStorage) -> Bottle:
             "target_name": target_name,
             "param_importances": [
                 {
-                    "name": i[0],
-                    "importance": i[1],
-                    "distribution": get_distribution_name(i[0], study),
+                    "name": name,
+                    "importance": importance,
+                    "distribution": get_distribution_name(name, study),
                 }
-                for i in importances.items()
+                for name, importance in importances.items()
             ],
         }
 
--- a/optuna_dashboard/app.py
+++ b/optuna_dashboard/app.py
@@ -8,10 +8,11 @@ import traceback
 from typing import Union, Dict, List, Optional, TypeVar, Callable, Any, cast
 
 from bottle import Bottle, BaseResponse, redirect, request, response, static_file
+import optuna
 from optuna.exceptions import DuplicatedStudyError
 from optuna.storages import BaseStorage
-from optuna.trial import FrozenTrial
-from optuna.study import StudyDirection, StudySummary
+from optuna.trial import FrozenTrial, TrialState
+from optuna.study import StudyDirection, StudySummary, Study
 
 from . import serializer
 
@@ -92,6 +93,13 @@ def get_trials(
         trials_last_fetched_at[study_id] = datetime.now()
         trials_cache[study_id] = trials
     return trials
+
+
+def get_distribution_name(param_name: str, study: Study) -> str:
+    for trial in study.trials:
+        if param_name in trial.distributions:
+            return trial.distributions[param_name].__class__.__name__
+    assert False
 
 
 def create_app(storage: BaseStorage) -> Bottle:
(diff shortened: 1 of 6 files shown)


c-bata:
Wow, great! I'll review tomorrow.


> There're two ways to add hyperparameter importance chart. One is to implement the evaluator which completely mirrors the one used by optuna via typescript and another is to use the optuna `optuna.importance.get_param_importances` API directly.
> 
> The former makes the logic cleaner but involves extra computation in the browser, and implementing a complete parameter importance function by ts requires an update whenever the optuna API changes.
> 
> The latter one is much more flexible and only requires introducing another API call. But the con is it takes longer for the server to get the importance result.
> 
> Here we take the second approach and only consider the single objective function case. For future multi-objective case, we can extend the API with query parameters representing params and target selected by the user.

I completely agree with your decision. It is quite difficult to port an importance evaluator into TypeScript because fANOVA evaluator requires the random forest regression model.

optuna_dashboard/app.py
```diff
@@ -94,6 +95,13 @@ def get_trials(
     return trials
 
 
+def get_distribution_name(param_name: str, study: Study) -> str:
+    for trial in study.trials:
+        if param_name in trial.distributions:
+            return trial.distributions[param_name].__class__.__name__
+    assert False
```
c-bata, line 102: Could you add a message?

```suggestion
    assert False, "Must not reach here."
```

optuna_dashboard/app.py
```diff
@@ -185,6 +193,38 @@ def get_study_detail(study_id: int) -> BottleViewReturn:
...
+        study_name = storage.get_study_name_from_id(study_id)
+        study = Study(study_name=study_name, storage=storage)
+
+        trials = [trial for trial in study.trials if trial.state == TrialState.COMPLETE]
+        if len(trials) == 0:
+            return ""
+        evaluator = None
+        params = None
+        target = None
+        importances = optuna.importance.get_param_importances(
```
c-bata, line 210: It seems that the default evaluator is fANOVA which depends on scikit-learn. So I think we need to add scikit-learn in `install_requires`, right?

If so, please merge main branch into `chenghuzi:plot_param_importances` and udpate `setup.cfg`.

optuna_dashboard/app.py
```diff
@@ -185,6 +193,38 @@ def get_study_detail(study_id: int) -> BottleViewReturn:
         trials = get_trials(storage, study_id)
         return serializer.serialize_study_detail(summary, trials)
 
+    @app.get("/api/studies/<study_id:int>/param_importances")
+    @handle_json_api_exception
+    def get_param_importances(study_id: int) -> BottleViewReturn:
+        # TODO: add support for selecting params and targets via query parameters.
```
c-bata, line 199:
```suggestion
        # TODO(chenghuzi): add support for selecting params and targets via query parameters.
```

optuna_dashboard/static/apiClient.ts
```diff
@@ -168,3 +168,23 @@ export const deleteStudyAPI = (studyId: number) => {
...
+    .get<ParamImportancesResponse>(
+      `/api/studies/${studyId}/param_importances`,
+      {}
+    )
+    .then((res) => {
+      return res.data
+    })
+}
+
+export default getParamImportances
```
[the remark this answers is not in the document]
chenghuzi: Yeah, sorry I forgot that part

optuna_dashboard/app.py
chenghuzi: Sure, will update in next commit

optuna_dashboard/static/components/HyperparameterImportances.tsx
```diff
@@ -0,0 +1,88 @@
...
+export const HyperparameterImportances: FC<{
+  studyId: number
+}> = ({ studyId }) => {
+  useEffect(() => {
+    async function fetchAndPlotParamImportances(studyId: number) {
+      const paramsImportanceData = await getParamImportances(studyId)
+      plotParamImportances(paramsImportanceData)
+    }
+    fetchAndPlotParamImportances(studyId)
+  })
```
c-bata, line 37: https://user-images.githubusercontent.com/5564044/111020398-d5698800-8408-11eb-9a3d-15f89e455eaa.mp4


Please set `deps` to avoid sending API requests at every render.

```suggestion
  }, [studyId])
```

optuna_dashboard/app.py
```diff
@@ -185,6 +193,38 @@ def get_study_detail(study_id: int) -> BottleViewReturn:
...
+
+        return {
+            "target_name": target_name,
+            "param_importances": [
+                {
+                    "name": i[0],
+                    "importance": i[1],
+                    "distribution": get_distribution_name(i[0], study),
+                }
+                for i in importances.items()
```
c-bata, line 224: [nits] A bit refactoring.

```suggestion
                    "name": name,
                    "importance": importance,
                    "distribution": get_distribution_name(name, study),
                }
                for name, importance in importances.items()
```

c-bata (changes requested):
Overall looks good and easy to read! I left some minor suggestions.

optuna_dashboard/static/components/HyperparameterImportances.tsx
chenghuzi: Oh thanks! I missed that part, re-rendering could be unnecessary.

chenghuzi: I was just thinking sometimes `studyId` could not be a proper dependency as trials may get updated while the `studyId ` remains the same while parameter importance requires a re-evaluation.

c-bata (approved):
LGTM!! For further improvements, we can work on following tasks in another pull requests:

* **In-memory cache**: `get_param_importances()` is computationally expensive  because of the training of RandomForest model. So it's better to cache the result at server-side like `trials` are cached.
* **Multi-objective support**: It seems that Optuna's importance module does not still support multi-objective study. So we need to send a pull request to Optuna for this.
* **Calculate hyperparameter importances on Web Assembly**: After I implement RandomForest and fANOVA algorithm in Goptuna, we can calculate hyperparameter importances at WebAssembly. It's a bit technically difficult, but interesting.

PR #55, opened 2021-03-11
Use setup.cfg for packaging

c-bata:

## Reference Issues/PRs

None

Use setup.cfg instead of setup.py.


setup.cfg
```diff
@@ -1,3 +1,53 @@
...
+    black
+    mypy
+    flake8
+
+[options.entry_points]
+console_scripts =
+    optuna-dashboard = optuna_dashboard.cli:main
+
+[options.package_data]
+optuna_dashboard = public/*
```
c-bata, line 45: memo: This seems to be ignored 🤔

c-bata: This is due to the following line.
https://github.com/pypa/setuptools/blob/fb7ab81a3d080422687bad71f9ae9d36eeefbee2/setuptools/command/sdist.py#L114-L121

(`build_py.data_files` holds the files in `public/bundle.js` and so on, but ignored because include_package_data is true.)

PR #70, opened 2021-04-07
Fix a bug to crash when trials are pruned with no intermediate values

### LINKED ISSUE #68: Dashboard is not rendered if trials are pruned without intermediate values
# Bug reports

This might be design, but I found that the dashboard page was not rendered if a study contained pruned trials without intermediate values.

## Expected Behavior

Studies with pruned trials having no intermediate values should be rendered. Or, an error page should be provided if they are not acceptable.

## Current Behavior and Steps to Reproduce

The following is an example code to reproduce the issue. In this example, I pruned trials whose objective values exceed a threshold. A list of studies were rendered as expected, but It showed blank page if I clicked the study name.

image

```python
import optuna


def objective(trial):
    # Binh and Korn function with constraints.
    x = trial.suggest_float("x", -15, 30)
    y = trial.suggest_float("y", -15, 30)

    v = x ** 2 + y ** 2

    # Prune if the objective values exceed the threshold.
    if v > 100:
        raise optuna.TrialPruned()

    return v


if __name__ == "__main__":
    study = optuna.create_study(
        storage="sqlite:///foo.db",
    )
    study.optimize(objective, n_trials=32, timeout=600)

    print("Number of finished trials: ", len(study.trials))
    print("Number of complete trials: ", len(study.get_trials(states=(optuna.trial.TrialState.COMPLETE,))))
```

## Context

Versions:
* optuna: 2.8.0.dev
* optuna-dashboard: 0.3.1

c-bata:

## Reference Issues/PRs

Fix #68 

When trials are pruned without intermediate values, `FrozenTrial.values` is None. It is the reason why the application will be crashed.

### DIFF
--- a/optuna_dashboard/static/components/GraphHistory.tsx
+++ b/optuna_dashboard/static/components/GraphHistory.tsx
@@ -187,7 +187,7 @@ const plotHistory = (
   }
 
   let filteredTrials = study.trials.filter(
-    (t) => t.state === "Complete" || t.state === "Pruned"
+    (t) => t.state === "Complete" || (t.state === "Pruned" && t.values && t.values.length > 0)
   )
   if (filterCompleteTrial) {
     filteredTrials = filteredTrials.filter((t) => t.state !== "Complete")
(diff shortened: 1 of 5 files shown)


c-bata:
ScreenShot 2021-04-07 13 17 41

PR #90, opened 2021-04-22
Show tick labels and add log scale in slice graph

simonhessner:

* [ x] I agree this patch may be ported to [Goptuna](https://github.com/c-bata/goptuna) by other Goptuna contributors.

## Reference Issues/PRs

No referenced issue as far as I know. I just found this a useful addition and I hope it helps other people too.

There are two additions in the slice graph:

1.  The x axis now shows tick labels (for both numerical and categorical types)
2.  It is now possible to switch between log scale and linear scale

### DIFF
--- a/optuna_dashboard/static/components/GraphSlice.tsx
+++ b/optuna_dashboard/static/components/GraphSlice.tsx
@@ -32,14 +32,15 @@ export const GraphSlice: FC<{
   const trials: Trial[] = study !== null ? study.trials : []
   const [objectiveId, setObjectiveId] = useState<number>(0)
   const [selected, setSelected] = useState<string | null>(null)
+  const [log, setLog] = useState<boolean>(false)
   const paramNames = study?.union_search_space.map((s) => s.name)
   if (selected === null && paramNames && paramNames.length > 0) {
     setSelected(paramNames[0])
   }
 
   useEffect(() => {
-    plotSlice(trials, objectiveId, selected)
-  }, [trials, objectiveId, selected])
+    plotSlice(trials, objectiveId, selected, log)
+  }, [trials, objectiveId, selected, log])
 
   const handleObjectiveChange = (
     event: React.ChangeEvent<{ value: unknown }>
@@ -92,7 +93,8 @@ export const GraphSlice: FC<{
 const plotSlice = (
   trials: Trial[],
   objectiveId: number,
-  selected: string | null
+  selected: string | null,
+  log: boolean
 ) => {
   if (document.getElementById(plotDomId) === null) {
     return
@@ -116,6 +118,7 @@ const plotSlice = (
     },
     yaxis: {
       title: "Objective Values",
+      type: log ? "log" : "linear",
       zerolinecolor: "#f2f5fa",
       zerolinewidth: 2,
       linecolor: "#f2f5fa",
@@ -171,6 +174,10 @@ const plotSlice = (
       linewidth: 5,
       gridcolor: "#f2f5fa",
       gridwidth: 1,
+      tickfont: {
+        "color": "#000000"
+      },
+      automargin: true  // Otherwise the label is outside of the plot
     }
     plotly.react(plotDomId, trace, layout)
   } else {
@@ -189,7 +196,7 @@ const plotSlice = (
         // xaxis: paramName,
         marker: {
           color: "#185799",
-        },
+        }
       },
     ]
     layout["xaxis"] = {
@@ -205,7 +212,7 @@ const plotSlice = (
       },
       tickvals: tickvals,
       ticktext: vocabArr,
-      automargin: true
+      automargin: true  // Otherwise the label is outside of the plot
     }
     plotly.react(plotDomId, trace, layout)
   }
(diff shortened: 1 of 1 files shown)


optuna_dashboard/static/components/GraphSlice.tsx
```diff
@@ -171,6 +188,10 @@ const plotSlice = (
       linewidth: 5,
       gridcolor: "#f2f5fa",
       gridwidth: 1,
+      tickfont: {
+        color: "#000000",
+      },
```
c-bata, line 193: [nits] Could you tell me what is this attribute for?

optuna_dashboard/static/components/GraphSlice.tsx
```diff
@@ -171,6 +188,10 @@ const plotSlice = (
...
+      automargin: true, // Otherwise the label is outside of the plot
```
c-bata, line 194: Thank you for adding this option. Looks good but I'm wondering why the label of xAxis is hidden when changed xAxis. Can you find the cause of this problem?

https://user-images.githubusercontent.com/5564044/115712873-4ad55a80-a3b0-11eb-97ab-2d3fdc6a9f06.mp4

c-bata (commented):
Basically looks good to me. I left some review comments.

optuna_dashboard/static/components/GraphSlice.tsx
simonhessner: This does not happen for me. Which browser are you using? I am using Chrome 90.0 on Windows (But server runs on Ubuntu)

Maybe it is due to a different database? Could you share the datebase with me so that I can look into it?

c-bata: Hmm... I see.

> Maybe it is due to a different database? Could you share the datebase with me so that I can look into it?

Of course, sure. I checked your pull request on the following program.

```python
import optuna
from optuna_dashboard.app import create_app


host = "127.0.0.1"
port = 8080


def create_optuna_storage() -> optuna.storages.InMemoryStorage:
    storage = optuna.storages.InMemoryStorage()

    # Single-objective study
    study = optuna.create_study(study_name="single-objective", storage=storage)

    def objective_single(trial: optuna.Trial) -> float:
        x1 = trial.suggest_float("x1", 0, 10)
        x2 = trial.suggest_float("x2", 0, 10)
        return (x1 - 2) ** 2 + (x2 - 5) ** 2

    study.optimize(objective_single, n_trials=100)
    return storage


def main() -> None:
    storage = create_optuna_storage()
    app = create_app(storage)
    app.run(host=host, port=port, reloader=True)


if __name__ == "__main__":
    main()
```

optuna_dashboard/static/components/GraphSlice.tsx
simonhessner: This makes sure that the color of the x labels is black. It might not be strictly required, but since it was present in the non-numerical case, I added it here for consistency. If you want I can remove it in both places.

simonhessner: By the way, there is a lot of duplicate code in the if(isnum) case and the else case. These cases basically only differ in the trace and the code could be simplyfied. I could do that if you want

simonhessner: I have another question about the code. On line 177 you have `xaxis: selected` and on line 210 you have `// xaxis: paramName`. So in one case (isnum) it is set and in the other case it is not. What is the effect of this parameter? I did not notice any difference when I commented it out, is it required?

c-bata: > This makes sure that the color of the x labels is black. It might not be strictly required, but since it was present in the non-numerical case, I added it here for consistency. If you want I can remove it in both places.

I see. You are right 👍 

> By the way, there is a lot of duplicate code in the if(isnum) case and the else case. These cases basically only differ in the trace and the code could be simplyfied. I could do that if you want

Sounds good! Welcome.

optuna_dashboard/static/components/GraphSlice.tsx
simonhessner: Thank you. So far I was using sqlite as a storage. I can't find an option in optuna to read from a in-memory storage. How did you do that?

simonhessner: wait, I am stupid...

simonhessner: Thank you. With this data I can reproduce the bug.

simonhessner: Fun fact: This bug was caused by the line that I was asking about in the other comment. Removing `xaxis: selected` solved it. I don't really know what this attribute was used for, but it looks like everything works now. I'll push the change in a second

optuna_dashboard/static/components/GraphSlice.tsx
simonhessner: I'll make these changes in a separate PR because I don't know when I have time. Is everything in the current PR ok?

c-bata: > I'll make these changes in a separate PR because I don't know when I have time. Is everything in the current PR ok?

Sure!

c-bata (approved):
Thank you for your swift responses! LGTM 👍

simonhessner:
Great, thank you! And thanks for this nice tool :-)

Issue #94, opened 2021-05-02
UI based unit test & functional test libraries setup


hegdeashwin:
# Feature Request

I will recommend to add Test suits/framework on UI layer for unit and functional testing during development.

As the dashboard is built using React, I will highly recommend to use the most popular test library [@testing-library/react](https://github.com/testing-library/react-testing-library)  and [jest](https://jestjs.io/docs/getting-started) to write unit tests for UI/Component layer and utility functions.

For functional tests, we could use [Cypress.io](https://www.cypress.io/).

I can start the contribution on the above work. Let me know your thoughts, I'm not sure if you guys have already planned this.

keisuke-umezawa:
Note
For functional tests, we introduced Playwright. ref: https://github.com/optuna/optuna-dashboard/pull/484.

PR #110, opened 2021-06-18
Add unit test for TrialTable

chenghuzi:

## Reference Issues/PRs

#94 

Add unit test for sorting in `TrialTable`

### DIFF
--- a/optuna_dashboard/static/components/DataGrid.tsx
+++ b/optuna_dashboard/static/components/DataGrid.tsx
@@ -54,7 +54,7 @@ interface DataGridColumn<T> {
   less?: (a: T, b: T) => number
   filterable?: boolean
   toCellValue?: (rowIndex: number) => string | React.ReactNode
-  padding?: "default" | "checkbox" | "none"
+  padding?: "normal" | "checkbox" | "none"
 }
 
 interface RowFilter<T> {
@@ -156,7 +156,7 @@ function DataGrid<T>(props: {
               {columns.map((column, index) => (
                 <TableCell
                   key={index}
-                  padding={column.padding || "default"}
+                  padding={column.padding || "normal"}
                   sortDirection={orderBy === column.field ? order : false}
                 >
                   <span className={classes.tableHeaderCell}>
@@ -225,8 +225,8 @@ function DataGrid<T>(props: {
         count={filteredRows.length}
         rowsPerPage={rowsPerPage}
         page={page}
-        onChangePage={handleChangePage}
-        onChangeRowsPerPage={handleChangeRowsPerPage}
+        onPageChange={handleChangePage}
+        onRowsPerPageChange={handleChangeRowsPerPage}
       />
     </div>
   )
@@ -273,7 +273,7 @@ function DataGridRow<T>(props: {
           return column.filterable ? (
             <TableCell
               key={`${row[keyField]}:${column.field}:${columnIndex}`}
-              padding={column.padding || "default"}
+              padding={column.padding || "normal"}
               onClick={(e) => {
                 handleClickFilterCell(column.field, row[column.field])
               }}
@@ -283,7 +283,7 @@ function DataGridRow<T>(props: {
           ) : (
             <TableCell
               key={`${row[keyField]}:${column.field}:${columnIndex}`}
-              padding={column.padding || "default"}
+              padding={column.padding || "normal"}
             >
               {cellItem}
             </TableCell>
--- a/package.json
+++ b/package.json
@@ -5,12 +5,13 @@
   "description": "Dashboard for Optuna",
   "main": "index.js",
   "scripts": {
-    "fmt": "prettier --write \"optuna_dashboard/static/**/*.{ts,tsx}\"",
-    "lint": "eslint . --ext .ts,.tsx && prettier --list-different \"optuna_dashboard/static/**/*.{ts,tsx}\"",
+    "fmt": "prettier --write \"optuna_dashboard/static/**/*.{ts,tsx}\" \"frontend_tests/*.{ts,tsx}\"",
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -27,14 +29,23 @@
...
+    "@babel/preset-react": "^7.14.5",
+    "@types/enzyme": "^3.10.8",
+    "@types/jest": "^26.0.23",
     "@types/plotly.js": "^1.54.10",
     "@types/react": "^17.0.9",
     "@types/react-dom": "^17.0.6",
     "@types/react-router-dom": "^5.1.7",
     "@typescript-eslint/eslint-plugin": "^4.26.1",
     "@typescript-eslint/parser": "^4.26.1",
+    "@wojtekmaj/enzyme-adapter-react-17": "^0.6.2",
```
c-bata, line 42: Can we avoid the use of this library? It looks unofficial and not so popular library.

package.json
```diff
@@ -27,14 +29,23 @@
...
+    "enzyme": "^3.11.0",
     "eslint": "^7.28.0",
+    "jest": "^27.0.4",
+    "jest-canvas-mock": "^2.3.1",
```
c-bata, line 46: Is this library required?

optuna_dashboard/static/components/DataGrid.tsx
```diff
@@ -313,7 +313,7 @@ function getComparator<T>(
     : (a, b) => -descendingComparator<T>(a, b, columns, orderBy)
 }
 
-function descendingComparator<T>(
+export function descendingComparator<T>(
```
c-bata, line 316: I think we should not export this function for unit tests.

tsconfig.json
```diff
@@ -24,7 +24,8 @@
     "./optuna_dashboard/static/index.tsx"
   ],
   "include": [
-    "./optuna_dashboard/static/types/**/*"
+    "./optuna_dashboard/static/types/**/*",
+    "./tests/frontend_tests"
```
c-bata, line 28: Could you put `frontend_tests` directory on the repository root?

(Let me rename an existing `tests` directory with `python_tests` in the following pull requests.)

tests/frontend_tests/TrialTable.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+  const trial_table_component = mount(<TrialTable studyDetail={study_detail} />)
+  const head_cells = trial_table_component
+    .find("ForwardRef(TableHead)")
+    .find("WithStyles(ForwardRef(TableCell))")
+
+  const columns: DataGridColumn<Trial>[] = trial_table_component
+    .find("DataGrid")
+    .prop("columns")
+
+  for (let i = 0; i < head_cells.length; i++) {
```
c-bata, line 94: These codes are known as one factor that makes tests more complicated than they really should be. Could you make tests more simple?
http://xunitpatterns.com/Conditional%20Test%20Logic.html

c-bata (changes requested):
Hi @chenghuzi. Thank you for your pull requests.
I put some review comments.

package.json
c-bata: I asked my friend who is a frontend engineer, if I should use Enzyme on optuna-dashboard. That's because Enzyme doesn't support React v17, that was released about six months ago.

He recommended me to use https://github.com/testing-library/react-testing-library instead of Enzyme. Enzyme hasn't been able to keep up with the release cycle of React.js. So could you remove Enzyme from the dependencies and use react-testing-library instead?

package.json
```diff
@@ -5,18 +5,20 @@
...
-    "build:prd": "NODE_ENV=production webpack"
+    "build:prd": "NODE_ENV=production webpack",
+    "test": "jest tests/frontend_tests"
   },
   "author": "Masashi Shibata",
   "license": "MIT",
   "dependencies": {
     "@material-ui/core": "^4.11.4",
     "@material-ui/icons": "^4.11.2",
+    "@types/enzyme-adapter-react-16": "^1.0.6",
```
c-bata, line 21: Please install these testing libraries as devDependencies.

package.json
chenghuzi: Yes, I was first using Enzyme as I didn't pull the latest change, which upgrades React from 16 to 17. Then I realized that Enzyme hasn't supported 17 yet. So I have to find some unofficial adapter to replace the older one. Don't worry, I'll check if using react-testing-library will be better (which I believe will be the case)

package.json
chenghuzi: Jest is used. jest-canvas-mock is also used as we have plotly.js in the dashboard, which requires this lib to setup a canvas mock for testing. See `jest.config.js` for the setup

optuna_dashboard/static/components/DataGrid.tsx
chenghuzi: I see, we can implement a similar but simplified one in the test file.

tsconfig.json
chenghuzi: No problem

package.json
chenghuzi: Just did some research on https://github.com/testing-library/react-testing-library and found that their support for `<table>` element is poor, so to select a whole column requires some hacky operations as react-testing-library does not encourage users to detect props change.

optuna_dashboard/static/components/DataGrid.tsx
```diff
@@ -54,7 +54,7 @@ interface DataGridColumn<T> {
   less?: (a: T, b: T) => number
   filterable?: boolean
   toCellValue?: (rowIndex: number) => string | React.ReactNode
-  padding?: "default" | "checkbox" | "none"
+  padding?: "normal" | "checkbox" | "none"
```
c-bata, line 57: Good catch 👍

c-bata (approved):
Looks simple and clear 💯  Perfect 😀 

We can work on adding tests for a filtering option at following pull requests.
LGTM!!

PR #115, opened 2021-06-29
Add incremental sync for the study summary API

chenghuzi:

Before:
Every time when you call API to update the study, the backend returns all trials, which may be redundant when the study is being updated frequently.

After:
You will only fetch all trials in the first call to study details API and all later calls will only add those updated trials.

### DIFF
--- a/optuna_dashboard/app.py
+++ b/optuna_dashboard/app.py
@@ -204,11 +204,12 @@ def create_app(storage: BaseStorage) -> Bottle:
     @handle_json_api_exception
     def get_study_detail(study_id: int) -> BottleViewReturn:
         response.content_type = "application/json"
+        ntrials_client = int(request.params["ntrials_client"])
         summary = get_study_summary(storage, study_id)
         if summary is None:
             response.status = 404  # Not found
             return {"reason": f"study_id={study_id} is not found"}
-        trials = get_trials(storage, study_id)
+        trials = get_trials(storage, study_id)[ntrials_client:]
         intersection, union = get_search_space(study_id, trials)
         return serializer.serialize_study_detail(summary, trials, intersection, union)
 
--- a/optuna_dashboard/app.py
+++ b/optuna_dashboard/app.py
@@ -204,12 +204,19 @@ def create_app(storage: BaseStorage) -> Bottle:
     @handle_json_api_exception
     def get_study_detail(study_id: int) -> BottleViewReturn:
         response.content_type = "application/json"
-        ntrials_client = int(request.params["ntrials_client"])
+        try:
+            before = int(request.params["before"])
+            assert before >= 0
+        except AssertionError:
+            response.status = 400  # Bad parameter
+            return {"reason": "`before` should be larger or equal 0."}
+        except KeyError:
+            before = 0
         summary = get_study_summary(storage, study_id)
         if summary is None:
             response.status = 404  # Not found
             return {"reason": f"study_id={study_id} is not found"}
-        trials = get_trials(storage, study_id)[ntrials_client:]
+        trials = get_trials(storage, study_id)[before:]
         intersection, union = get_search_space(study_id, trials)
         return serializer.serialize_study_detail(summary, trials, intersection, union)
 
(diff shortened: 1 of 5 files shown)


optuna_dashboard/app.py
```diff
@@ -204,11 +204,12 @@ def delete_study(study_id: int) -> BottleViewReturn:
     @handle_json_api_exception
     def get_study_detail(study_id: int) -> BottleViewReturn:
         response.content_type = "application/json"
+        ntrials_client = int(request.params["ntrials_client"])
```
c-bata, line 207: It will raise an exception if clients call the API without this query parameter. Please return all trials when given no query parameter.

c-bata: Please add unit tests for the following scenarios.

1. Call the API without query parameters.
1. Call the API with invalid query parameters (e.g. It should return BAD REQUEST when given a large trial number such as 1000 even though there are only 100 trials).
1. Call the API with valid query parameters.

optuna_dashboard/static/action.ts
```diff
@@ -33,8 +33,14 @@ export const actionCreator = () => {
   }
 
   const updateStudyDetail = (studyId: number) => {
-    getStudyDetailAPI(studyId)
+    getStudyDetailAPI(
+      studyId,
+      studyId in studyDetails ? studyDetails[studyId].trials.length : 0
```
c-bata, line 38: `RUNNING` and `WAITING` trials will be updated. So you need to pass the last completed trial number.

c-bata (changes requested):

optuna_dashboard/app.py
c-bata: And could you rename query parameters to `before`? I think it is the common naming in the cursor-based pagination for the Web API.

c-bata (changes requested):

optuna_dashboard/app.py
chenghuzi: No problem. These have been resolved in the latest commit

optuna_dashboard/app.py
```diff
@@ -204,11 +204,19 @@ def delete_study(study_id: int) -> BottleViewReturn:
     @handle_json_api_exception
     def get_study_detail(study_id: int) -> BottleViewReturn:
         response.content_type = "application/json"
+        try:
+            before = int(request.params["before"])
+            assert before >= 0
```
c-bata, line 209: @chenghuzi I'm sorry. We should name this `after` because we want to get trials after the given trial number.

tests/test_api.py
```diff
@@ -25,6 +25,59 @@ def test_get_study_summaries(self) -> None:
         study_summaries = json.loads(body)["study_summaries"]
         self.assertEqual(len(study_summaries), 2)
 
+    def test_get_study_details(self) -> None:
```
c-bata, line 28: Could you separate test cases to avoid [Assertion Roulette](http://xunitpatterns.com/Assertion%20Roulette.html)?

c-bata (changes requested):

optuna_dashboard/static/action.ts
```diff
@@ -33,8 +33,23 @@ export const actionCreator = () => {
...
+    let nLocalFixedTrials = 0
+    if (studyId in studyDetails) {
+      for (const trial of studyDetails[studyId].trials) {
+        if (!["Running", "Waiting"].includes(trial.state)) {
+          nLocalFixedTrials += 1
+        } else {
+          break
+        }
+      }
+    }
```
c-bata, line 45: You may be able to simplify this by using `array.findIndex()`.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex

optuna_dashboard/app.py
chenghuzi: Fixed in the latest commit

optuna_dashboard/static/components/StudyDetail.tsx
```diff
@@ -117,7 +117,7 @@ export const StudyDetail: FC = () => {
       action.updateStudyDetail(studyIdNumber)
     }, reloadInterval * 1000)
     return () => clearInterval(intervalId)
-  }, [reloadInterval])
+  }, [reloadInterval, studyDetail])
```
c-bata, line 120: Could you add a todo comment like below?

```
// TODO(chenghuzi): Reduce the number of calls to setInterval and clearInterval.
```

c-bata (approved):
LGTM with a nit change request.

PR #217, opened 2022-04-24
Add cached extra study property

yoshinobc:

### Motivation
Eliminate unnecessary calculations when retrieving cached study properties.

### Description of the changes
- Add call to `get_chached_extra_study_property` method in `_app.py`
- Integration of `_search_space.py` and `_intermediate_values.py`. Update both information in a single update.


optuna_dashboard/_cached_extra_study_property.py
```diff
@@ -54,6 +52,9 @@ def union(self) -> SearchSpaceListT:
         return union
 
     def update(self, trials: List[FrozenTrial]) -> None:
+        if self.has_intermediate_values:
+            return
```
c-bata, line 56: Even if `self.has_intermediate_values` is False, we need to continue for-loop to update the search space cache.

optuna_dashboard/_cached_extra_study_property.py
```diff
@@ -65,11 +72,15 @@ def update(self, trials: List[FrozenTrial]) -> None:
             if trial.state not in states_of_interest:
                 continue
 
+            if len(trial.intermediate_values) > 0:
```
c-bata, line 75: [nits] How about changing here like following to reduce a bit of computation since we don't need to update `self.has_intermediate_values` if it's already True?

```suggestion
            if self.has_intermediate_values or len(trial.intermediate_values) > 0:
```

python_tests/test_cached_extra_study_property.py
```diff
@@ -0,0 +1,177 @@
...
+        ]
+        cached_extra_study_property = _CachedExtraStudyProperty()
+        cached_extra_study_property.update(trials)
+        self.assertTrue(cached_extra_study_property.has_intermediate_values)
+
+    def test_no_trials(self) -> None:
+        trials = []
+        cached_extra_study_property = _CachedExtraStudyProperty()
+        cached_extra_study_property.update(trials)
+        self.assertFalse(cached_extra_study_property.has_intermediate_values)
```
c-bata, line 177: How about splitting these test cases into another TestCase class? This test class looks somewhat large. These logics are defined in the same class (`_CachedExtraStudyProperty`) though.

c-bata (commented):
Hi @yoshinobc.
Thank you for your pull request!

You changes looks almost good after solving lint errors and my suggestions.

optuna_dashboard/_cached_extra_study_property.py
yoshinobc: Thank you very much.
When ` self.has_intermediate_values` is True , there is no need to run L76, so
Isn't this place `if not self.has_intermediate_values and len(trial.intermediate_values) > 0:`

c-bata: Oh, sorry. My suggestion is incorrect.

We can actually skip the evaluation for `len(trial.intermediate_values) > 0` if making the condition of if statement like `not self.has_intermediate_values or len(trial.intermediate_values) > 0`.

c-bata (approved):
LGTM after passed all CI tests!

c-bata:
Merged! Thank you for your contribution 👍

PR #226, opened 2022-04-29
Fix a bug when handling categorical parameter that contain number in the string

### LINKED ISSUE #198: Strings not embedded properly in Parallel Coordinate plot.
### Description

TL;DR discrete hyperparameters with numbers in the string are not represented correctly in the parallel coordinate plots.
 
I have a hyperparameter `configuration` which regulates the configuration of a MLP which looks like this "-1:512:64:512:-1", because the `suggest_categorical` does not support tuples or lists.

In the Parallel coordinate plot, this string is represented by `int: -1` and thus all configurations are mapped to the same y coordinate.

### How to Reproduce

Start an optimization with a hyper-parameter:
`suggest_categorical('hp1', ["-1:512:64:512:-1",  "-1:512:16:512:-1"])`

The parallel coordinate plot will show both configurations as 1 point.

This problem can be circumvented by using ["A:-1:512:64:512:-1",  "B:-1:512:16:512:-1"]

### Python version

3.9

### Optuna version

optuna 2.10.0

optuna-dashboard 0.6.1

### Web browser

Firefox

yoshinobc:

## Reference Issues/PRs
Fixes #198 

### Motivation
Fix a bug when handling categorical hyperparameters that contain number in the string.

The `parseFloat()` function used in isnum starts from the beginning and returns the parsed value when it is no longer a number.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseFloat#:~:text=A%20second%20decimal%20point%20also%20stops%20parsing%20(characters%20up%20to%20that%20point%20will%20still%20be%20parsed).
Therefore, a categorical parameter whose first letter is a number is treated as a numeric parameter (`isnum is true`).
`Number()` is now used to determine whether a number is a number or not.
スクリーンショット 2022-04-29 15 53 03

### Description of changes
- Change from 'parseFloat` to `Number` in `GraphParallelCorrdinate.tsx`, `GraphSlice.tsx`

スクリーンショット 2022-04-29 16 04 15
スクリーンショット 2022-04-29 16 04 04

### DIFF
--- a/optuna_dashboard/ts/components/GraphParallelCoordinate.tsx
+++ b/optuna_dashboard/ts/components/GraphParallelCoordinate.tsx
@@ -136,7 +136,7 @@ const plotCoordinate = (
       return param!.value
     })
     const isnum = valueStrings.every((v) => {
-      return !isNaN(parseFloat(v))
+      return !isNaN(Number(v))
     })
     if (isnum) {
       const values: number[] = valueStrings.map((v) => parseFloat(v))
--- a/optuna_dashboard/ts/components/GraphSlice.tsx
+++ b/optuna_dashboard/ts/components/GraphSlice.tsx
@@ -186,7 +186,7 @@ const plotSlice = (
   })
 
   const isnum = valueStrings.every((v) => {
-    return !isNaN(parseFloat(v))
+    return !isNaN(Number(v))
   })
   if (isnum) {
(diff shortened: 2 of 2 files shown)


c-bata (approved):
Looks Great To Me 💯 Thank you for your contribution!

PR #355, opened 2023-01-09
Add trial tell feature

keisuke-umezawa:

## Reference Issues/PRs

N/A

スクリーンショット 2023-01-09 21 10 13

- Add tell python API in the server side
- Call the above API in the client side
- It also applicable for multi-objectives
- To pass unittest, I need to update jest and test files.


c-bata (changes requested):
Thank you for the pull request. I left some comments.

optuna_dashboard/ts/action.ts
```diff
@@ -265,6 +266,34 @@ export const actionCreator = () => {
...
+  ) => {
+    const message =
+      values === undefined
+        ? `id=${trialId}, state=${state}`
+        : `id=${trialId}, state=${state}, values=${values}`
+    tellTrialAPI(studyId, trialId, state, values)
+      .then(() => {
+        enqueueSnackbar(`Success to update trial (${message})`, {
+          variant: "success",
+        })
```
c-bata, line 283: Could you update the Recoil's `studyDetailsState` here like `saveStudyNote` action does?
https://github.com/optuna/optuna-dashboard/blob/29a36bdb6affbb3b065b4acf588f0fee45acd779/optuna_dashboard/ts/action.ts#L196-L205

optuna_dashboard/ts/apiClient.ts
```diff
@@ -217,6 +217,24 @@ export const saveTrialNoteAPI = (
     })
 }
 
+export const tellTrialAPI = (
+  studyId: number,
+  trialId: number,
+  state: TrialState,
+  values?: string[]
+): Promise<void> => {
+  const req: { [name: string]: TrialState | string[] } = { state: state }
```
c-bata, line 226: Could you type the request like the following?

```
{
    state: TrialState
    values?: number[]
}
```

c-bata (changes requested):

optuna_dashboard/ts/components/TrialTable.tsx
```diff
@@ -293,6 +332,41 @@ export const TrialTable: FC<{
...
+                  variant="outlined"
+                  color="error"
+                  onClick={handleFailTrial}
+                >
+                  Fail trial
+                </Button>
+              </Box>
+            </form>
+          </Box>
+        </Grid>
```
hvy, line 369: Sorry I'm not very familiar with react but will this component always be shown even for completed trials? It'd be nice if we could hide the component for finished trials. What do you think?

c-bata: That sounds reasonable to me +1

optuna_dashboard/_app.py
```diff
@@ -411,6 +412,47 @@ def save_study_note(study_id: int) -> BottleViewReturn:
...
+        try:
+            study_name = storage.get_study_name_from_id(study_id)
+        except KeyError:
+            response.status = 404  # Not found
+            return {"reason": f"study_id={study_id} is not found"}
+
+        study = optuna.load_study(storage=storage, study_name=study_name)
+
+        try:
+            study.tell(trial_id, values=values, state=state)
```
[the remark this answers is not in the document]
hvy: Would you mind motivating this change a bit? 🙏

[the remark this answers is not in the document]
c-bata: Since the `Study` class is designed for Optuna users, some code may be unnecessary or inefficient in Optuna Dashboard.

For example, the following code is inefficient since it cannot use the trial cache that Optuna Dashboard has.
https://github.com/optuna/optuna/blob/master/optuna/study/_tell.py#L103

Also, the input values are validated in `study.tell()` function, but they need to be validated again in order for Optuna Dashboard's API server to return the proper response status and error messages.
https://github.com/optuna/optuna/blob/master/optuna/study/_tell.py#L127-L158

package.json
```diff
@@ -22,6 +22,7 @@
     "@mui/lab": "^5.0.0-alpha.114",
     "@mui/material": "^5.11.2",
     "axios": "^1.2.1",
+    "jest-environment-jsdom": "^29.3.1",
```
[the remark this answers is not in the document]
keisuke-umezawa: From some version of the jest, it requires "jest-environment-jsdom" additionally. If we do not have the library, it shows the following error.

```
> jest typescript_tests 
                                                           
● Validation Error:                                                                                                   
                                                                                                                      
  Test environment jest-environment-jsdom cannot be found. Make sure the testEnvironment configuration option points t
o an existing node module.                                                                                            
                                                                                                                      
  Configuration Documentation:                             
  https://jestjs.io/docs/configuration
                                                                                                                                                                             
As of Jest 28 "jest-environment-jsdom" is no longer shipped by default, make sure to install it separately.  
``` 

However, it should be installed only for the dev envirionment. So, I fixed it.

optuna_dashboard/ts/components/TrialTable.tsx
```diff
@@ -263,6 +273,35 @@ export const TrialTable: FC<{
   ]
 
   const collapseBody = (index: number) => {
+    const objectiveValuesLength = studyDetail?.directions.length
+    const [objectiveValues, setObjectiveValues] = useState(
+      Array(objectiveValuesLength).fill("")
+    )
```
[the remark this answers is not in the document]
keisuke-umezawa: As we discussed in the in-person meeting, it will be resolved in another PR.

optuna_dashboard/_app.py
```diff
@@ -411,6 +412,47 @@ def save_study_note(study_id: int) -> BottleViewReturn:
...
+        s = request.json.get("state", None)
+        vs = request.json.get("values", None)
+
+        try:
+            values = [float(v) for v in vs] if vs is not None else vs
+        except ValueError:
+            response.status = 400  # Bad request
+            return {"reason": "You need to pass float castable values"}
+
+        string2State = {
```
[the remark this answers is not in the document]
keisuke-umezawa: Fixed.

optuna_dashboard/_app.py
```diff
@@ -411,6 +412,47 @@ def save_study_note(study_id: int) -> BottleViewReturn:
...
+            "Running": TrialState.RUNNING,
```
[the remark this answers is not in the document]
keisuke-umezawa: Fixed.

optuna_dashboard/_app.py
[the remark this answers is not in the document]
keisuke-umezawa: To use `storage.get_trial_state_values`, it also requires some more lines. For example, https://github.com/optuna/optuna/blob/master/optuna/study/_tell.py#L168-L173 . In my opinion, it could be safer to rely on the `study.tell()` function because the optuna may add more code into the tell functionality.

optuna_dashboard/ts/action.ts
keisuke-umezawa: Fixed.

optuna_dashboard/ts/apiClient.ts
keisuke-umezawa: Fixed.

optuna_dashboard/_app.py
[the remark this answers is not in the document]
c-bata: Regarding the lines you commented, optuna-dashboard does not require them. I'm not sure if `study.tell()` will contain the code that Optuna Dashboard would require in the future, but at this time there is no such code. Rather, it contains code that is clearly unnecessary or inefficient. I would say we should not rely on `study.tell()` since it is designed for Optuna users, not an Optuna Dashboard.

What do you think?

optuna_dashboard/ts/components/TrialTable.tsx
keisuke-umezawa: Fixed.

keisuke-umezawa:
@c-bata @hvy 
Thank you for your reviews! Basically, I fixed the code as you suggested. Could you review it again?

optuna_dashboard/ts/action.ts
```diff
@@ -265,6 +285,44 @@ export const actionCreator = () => {
       })
   }
 
+  const tellTrial = (
+    studyId: number,
+    trialId: number,
+    state: TrialState,
+    values?: string[]
```
c-bata, line 292: To increase the reusability, could you parse values before this function calls? For example, if we add the feature to specify objective values with a slider or radio button, this action would be called without parsing the string.

```suggestion
    values?: number[]
```

typescript_tests/TrialTable.test.tsx
```diff
@@ -124,7 +126,11 @@ const studyDetail: StudyDetail = {
 
 it("Sort TrialTable by trial number", () => {
   const { getAllByRole, getByText } = render(
-    <TrialTable studyDetail={studyDetail} isBeta={false} />
+    <RecoilRoot>
+      <SnackbarProvider>
```
c-bata, line 130: Just a question, if we remove these two components here, will test cases fail?

optuna_dashboard/ts/components/TrialTable.tsx
```diff
@@ -293,6 +333,47 @@ export const TrialTable: FC<{
...
+                        id={`objective-${i}`}
+                        key={`objective-${i}`}
+                        label={`Objective ${i}`}
+                        type="number"
+                        value={objectiveValues[i]}
+                        onChange={(e: ChangeEvent<HTMLInputElement>) =>
+                          handleChangeValue(i, e)
+                        }
+                      />
+                    ))}
```
c-bata, line 356: There are multiple TextFields if the study is multi-objective. Could you add form labels to indicate which values should be entered in each field?

Also we can refer `studyDetail.objectiveNames` like the following.
https://github.com/optuna/optuna-dashboard/blob/29233909a2d8ecba4b6e1a16f0cfdb9d76091d74/optuna_dashboard/ts/components/GraphParetoFront.tsx#L62-L64

c-bata (changes requested):
Thanks for the update! I left some suggestions.

optuna_dashboard/ts/components/TrialTable.tsx
```diff
@@ -293,6 +333,47 @@ export const TrialTable: FC<{
...
+                Trial tell operations
+              </Typography>
+              <form onSubmit={handleSubmit}>
+                <Box margin={1}>
+                  <Stack direction="row" spacing={1}>
+                    {objectiveValues.map((value, i) => (
+                      <TextField
+                        id={`objective-${i}`}
+                        key={`objective-${i}`}
+                        label={`Objective ${i}`}
```
c-bata, line 349: How about displaying the `studyDetail.objectiveNames` defined by `set_objective_names()`? You can easily support implement it like the following code.

https://github.com/optuna/optuna-dashboard/blob/29233909a2d8ecba4b6e1a16f0cfdb9d76091d74/optuna_dashboard/ts/components/GraphParetoFront.tsx#L62-L64

optuna_dashboard/ts/components/TrialTable.tsx
hvy: Seemed good, let me resolve this comment. Thanks 🙇

optuna_dashboard/ts/components/TrialTable.tsx
```diff
@@ -263,6 +274,35 @@ export const TrialTable: FC<{
...
+      Array(objectiveValuesLength).fill("")
+    )
+    const handleSubmit = (e: FormEvent<HTMLFormElement>): void => {
+      e.preventDefault()
+      const studyId = (studyDetail as StudyDetail).id
+      const trialId = trials[index].number
+      action.tellTrial(
+        studyId,
+        trialId,
+        "Complete" as TrialState,
```
c-bata, line 288:
```suggestion
        "Complete",
```

optuna_dashboard/ts/components/TrialTable.tsx
```diff
@@ -263,6 +274,35 @@ export const TrialTable: FC<{
...
+      e: ChangeEvent<HTMLInputElement>
+    ): void => {
+      const newValues = [...objectiveValues]
+      newValues[index] = e.target.value
+      setObjectiveValues(newValues)
+    }
+    const handleFailTrial = (e: MouseEvent<HTMLButtonElement>): void => {
+      const studyId = (studyDetail as StudyDetail).id
+      const trialId = trials[index].number
+      action.tellTrial(studyId, trialId, "Fail" as TrialState)
```
c-bata, line 303:
```suggestion
      action.tellTrial(studyId, trialId, "Fail")
```

optuna_dashboard/ts/components/TrialTable.tsx
```diff
@@ -293,6 +333,47 @@ export const TrialTable: FC<{
             />
           </Box>
         </Grid>
+        {trials[index].state === ("Running" as TrialState) ? (
```
c-bata, line 336:
```suggestion
        {trials[index].state === "Running" ? (
```

c-bata (changes requested):
You should not cast string literals to TrialState type since it makes difficult to find typos.

optuna_dashboard/ts/components/TrialTable.tsx
```diff
@@ -293,6 +333,47 @@ export const TrialTable: FC<{
...
+              </Typography>
+              <form onSubmit={handleSubmit}>
+                <Box margin={1}>
+                  <Stack direction="row" spacing={1}>
+                    {objectiveValues.map((value, i) => (
+                      <TextField
+                        id={`objective-${i}`}
+                        key={`objective-${i}`}
+                        label={`Objective ${i}`}
+                        type="number"
```
c-bata, line 350: Floating point numbers seems to be prohibited when using `type="number"`. You need to set `step` or `pattern` attribute as well.
Screenshot 2023-01-16 14 30 54

optuna_dashboard/_app.py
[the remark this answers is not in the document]
hvy: Thanks for the clarification, and the discussion. 

Personally, it's fine to use the storage API directly, at least in this case since it's not a private method, and I understand that the dashboard calls "Study.tell" under different circumstances, for instance doesn't allow specifying the state. I'm slightly skeptical about having validation logic at several places, though, more than the speed gained by using the cache, since the study also checks for NaN, the number of objectives in the case of MO, etc. If we'd in the future allow other states such as pruning, then I'd suggest reconsidering using `Study.tell` here.

optuna_dashboard/ts/components/TrialTable.tsx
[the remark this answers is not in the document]
c-bata: Could you cherry-pick the above commit to resolve this comment?
https://github.com/optuna/optuna-dashboard/commit/ec5670507604060e07f3d7f85dab01bb68e02f52

hvy:
(Optional) Graying out the `SUBMIT` button when values are not entered (or even invalid) would be really great. Really optional.

optuna_dashboard/ts/components/TrialTable.tsx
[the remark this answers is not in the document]
keisuke-umezawa: Thank you! Fixed.

typescript_tests/TrialTable.test.tsx
keisuke-umezawa: Yes. In the TrialTable, we started using recoil and snackbar because we call a `action.tellTrial` inside. That is the reason I need to wrap it with `<RecoilRoot>` and `<SnackbarProvider>`.

optuna_dashboard/ts/components/TrialTable.tsx
keisuke-umezawa: Fixed.

optuna_dashboard/ts/components/TrialTable.tsx
keisuke-umezawa: Fixed.

optuna_dashboard/ts/components/TrialTable.tsx
keisuke-umezawa: Fixed.

optuna_dashboard/ts/components/TrialTable.tsx
keisuke-umezawa: I used the following one and it was fixed.

```
    type="text"
    pattern="^([1-9]\d*|0)(\.\d+)?$"
```

optuna_dashboard/ts/action.ts
keisuke-umezawa: Fixed.

optuna_dashboard/ts/components/TrialTable.tsx
keisuke-umezawa: Fixed.

optuna_dashboard/_app.py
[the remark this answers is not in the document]
keisuke-umezawa: Ok. I understand. I followed it.

keisuke-umezawa:
> (Optional) Graying out the SUBMIT button when values are not entered (or even invalid) would be great. Really optional.
Currently, values are validated by the textfield setting in the client side. So, even if the submit button is pushed when values are invalid or not entered, the values will be validated and blocked in the client side.

keisuke-umezawa:
@c-bata @hvy I followed your comments and fixed code. Could you check this PR again? Thank you.

optuna_dashboard/ts/components/TrialTable.tsx
```diff
@@ -263,6 +274,28 @@ export const TrialTable: FC<{
...
+      createRef<HTMLInputElement>()
+    )
+    const handleSubmit = (e: FormEvent<HTMLFormElement>): void => {
+      if (objectiveFormRefs === undefined) {
+        return
+      }
+
+      e.preventDefault()
+      const studyId = (studyDetail as StudyDetail).id
+      const trialId = trials[index].number
```
c-bata, line 287:
```suggestion
      const trialId = trials[index].trial_id
```

optuna_dashboard/ts/components/TrialTable.tsx
```diff
@@ -263,6 +274,28 @@ export const TrialTable: FC<{
...
+      const objectiveValues = objectiveFormRefs.map((ref) =>
+        ref.current ? Number(ref.current.value) : NaN
+      )
+      action.tellTrial(studyId, trialId, "Complete", objectiveValues)
+    }
+    const handleFailTrial = (e: MouseEvent<HTMLButtonElement>): void => {
+      const studyId = (studyDetail as StudyDetail).id
+      const trialId = trials[index].number
```
c-bata, line 295: ditto.

```suggestion
      const trialId = trials[index].trial_id
```

optuna_dashboard/ts/components/TrialTable.tsx
```diff
@@ -263,6 +274,28 @@ export const TrialTable: FC<{
...
+      e.preventDefault()
+      const studyId = (studyDetail as StudyDetail).id
+      const trialId = trials[index].number
+      const objectiveValues = objectiveFormRefs.map((ref) =>
+        ref.current ? Number(ref.current.value) : NaN
+      )
+      action.tellTrial(studyId, trialId, "Complete", objectiveValues)
+    }
+    const handleFailTrial = (e: MouseEvent<HTMLButtonElement>): void => {
+      const studyId = (studyDetail as StudyDetail).id
```
c-bata, line 294: ditto.

optuna_dashboard/ts/components/TrialTable.tsx
```diff
@@ -263,6 +274,28 @@ export const TrialTable: FC<{
...
+    const objectiveFormRefs = studyDetail?.directions.map((d) =>
+      createRef<HTMLInputElement>()
+    )
+    const handleSubmit = (e: FormEvent<HTMLFormElement>): void => {
+      if (objectiveFormRefs === undefined) {
+        return
+      }
+
+      e.preventDefault()
+      const studyId = (studyDetail as StudyDetail).id
```
c-bata, line 286: You should not cast `StudyDetail | null` to `StudyDetail`. It's not safe. Please return this function if `studyDetail === null`.

optuna_dashboard/_app.py
```diff
@@ -411,6 +412,45 @@ def save_study_note(study_id: int) -> BottleViewReturn:
...
+            return {"reason": f"You passed {s} as a state, which is not defined in Optuna."}
+        state = str_to_state[s]
+
+        try:
+            study_name = storage.get_study_name_from_id(study_id)
+        except KeyError:
+            response.status = 404  # Not found
+            return {"reason": f"study_id={study_id} is not found"}
+
+        study = optuna.load_study(storage=storage, study_name=study_name)
```
c-bata, line 443: You don't need to create `Study` class. Could you use `storage` directly.

optuna_dashboard/_app.py
```diff
@@ -411,6 +412,45 @@ def save_study_note(study_id: int) -> BottleViewReturn:
...
+    @json_api_view
+    def tell_trial(study_id: int, trial_id: int) -> BottleViewReturn:
+        s = request.json.get("state", None)
+        vs = request.json.get("values", None)
+
+        try:
+            values = [float(v) for v in vs] if vs is not None else vs
+        except ValueError:
+            response.status = 400  # Bad request
+            return {"reason": "You need to pass float castable values"}
```
c-bata, line 425: Please return 400 response when `state == "Complete" and values is None` as well.

optuna_dashboard/ts/action.ts
```diff
@@ -265,6 +285,44 @@ export const actionCreator = () => {
...
+        const index = studyDetails[studyId].trials.findIndex(
+          (t) => t.trial_id === trialId
+        )
+        if (index === -1) {
+          enqueueSnackbar(`Unexpected error happens. Please reload the page.`, {
+            variant: "error",
+          })
+          return
+        }
+        setTrialState(studyId, index, state, values?.map(Number))
```
c-bata, line 309: `values` has been already converted.

```suggestion
        setTrialState(studyId, index, state, values)
```

optuna_dashboard/ts/action.ts
```diff
@@ -265,6 +285,44 @@ export const actionCreator = () => {
       })
   }
 
+  const tellTrial = (
```
c-bata, line 288: [nit] It's a matter of preference, so please feel free to ignore this suggestion.

I prefer to split this action to avoid `values?: number[]`.

* `make_trial_complete(studyId: number, trialId: number, values: number[])`
* `make_trial_fail(studyId: number, trialId: number)`

optuna_dashboard/ts/components/TrialTable.tsx
```diff
@@ -263,6 +274,28 @@ export const TrialTable: FC<{
...
+      const trialId = trials[index].number
+      const objectiveValues = objectiveFormRefs.map((ref) =>
+        ref.current ? Number(ref.current.value) : NaN
```
c-bata, line 289: I think we should not call APIs when `ref.current` is undefined or null. It's an implicit behavior that may cause the critical bug. What do you think?

c-bata (changes requested):
Thank you for the update! I left some review comments.

optuna_dashboard/ts/components/TrialTable.tsx
keisuke-umezawa: Fixed

optuna_dashboard/ts/components/TrialTable.tsx
keisuke-umezawa: Fixed

optuna_dashboard/_app.py
keisuke-umezawa: Fixed.

optuna_dashboard/_app.py
keisuke-umezawa: Fixed

optuna_dashboard/ts/components/TrialTable.tsx
keisuke-umezawa: Thank you. I changed it as suggedsted.

optuna_dashboard/ts/action.ts
keisuke-umezawa: Fixed.

optuna_dashboard/_app.py
```diff
@@ -411,6 +412,45 @@ def save_study_note(study_id: int) -> BottleViewReturn:
         response.status = 204  # No content
         return {}
 
+    @app.post("/api/studies/<trial_id:int>/tell")
```
c-bata, line 415: How about renaming `studies` with `trials`?

```suggestion
    @app.post("/api/trials/<trial_id:int>/tell")
```

optuna_dashboard/_app.py
```diff
@@ -411,6 +412,45 @@ def save_study_note(study_id: int) -> BottleViewReturn:
...
+            }
+        state = str_to_state[s]
+
+        try:
+            storage.set_trial_state_values(trial_id, state, values)
+        except Exception as e:
+            response.status = 400  # Bad request
+            return {"reason": e.args}
+
+        response.status = 201
```
c-bata, line 451: You should return 204 here since the response does not contain any information.

```suggestion
        response.status = 204
```

optuna_dashboard/_app.py
```diff
@@ -411,6 +412,45 @@ def save_study_note(study_id: int) -> BottleViewReturn:
...
+            return {
+                "reason": "When you passed 'Complete' as a state, you also need to specify values."
+            }
+        state = str_to_state[s]
+
+        try:
+            storage.set_trial_state_values(trial_id, state, values)
+        except Exception as e:
+            response.status = 400  # Bad request
+            return {"reason": e.args}
```
c-bata, line 449: 500 should be returned here.

```suggestion
            response.status = 500
            return {"reason": f"Internal server error: {e}"}
```

optuna_dashboard/ts/apiClient.ts
```diff
@@ -217,6 +217,23 @@ export const saveTrialNoteAPI = (
...
+  state: TrialStateFinished,
+  values?: number[]
+): Promise<void> => {
+  const req: { state: TrialState; values?: number[] } = {
+    state: state,
+    values: values,
+  }
+
+  return axiosInstance
+    .post<void>(`/api/studies/${trialId}/tell`, req)
```
c-bata, line 231:
```suggestion
    .post<void>(`/api/trials/${trialId}/tell`, req)
```

optuna_dashboard/_app.py
```diff
@@ -411,6 +412,45 @@ def save_study_note(study_id: int) -> BottleViewReturn:
...
+            return {
+                "reason": f"You passed {s} as a state, but only 'Complete', 'Pruned' and 'Fail'"
+                "are acceptable states."
+            }
+        if s == "Complete" and values is None:
+            response.status = 400  # Bad request
+            return {
+                "reason": "When you passed 'Complete' as a state, you also need to specify values."
+            }
+        state = str_to_state[s]
```
c-bata, line 443: How about replacing these logic like the following?

https://github.com/optuna/optuna-dashboard/commit/dc0fdb6bddf22732402e4b48d777f9ab295077a5

In the above commit, I made following improvements.

* When given `{'state': 'Complete', 'values': 0}`, this view function raise `TypeError` exception. So we must handle it.
* I would say we can convert str to state and then check `state.is_finished()` to validate the state.

c-bata (changes requested):
Looks almost good to me. I leave some suggestions on JSON API implementations though.

optuna_dashboard/_app.py
keisuke-umezawa: Thank you 🙇  I cherry-picked it.

optuna_dashboard/_app.py
```diff
@@ -411,6 +419,34 @@ def save_study_note(study_id: int) -> BottleViewReturn:
         response.status = 204  # No content
         return {}
 
+    @app.post("/api/trials/<trial_id:int>/tell")
+    @json_api_view
+    def tell_trial(trial_id: int) -> BottleViewReturn:
+        state = str_to_trial_state.get(request.json.get("state"))
+        if state is None or not state.is_finished():
+            response.status = 400  # Bad request
+            return {"reason": "state must be either 'Complete', 'Pruned' or 'Fail'."}
```
hvy, line 428: If we don't need pruning for the time being, we could choose to not support it, i.e. return early. The logic is quite different from `Study.tell` in Optuna, and it'd make tests simpler. Just an idea.

optuna_dashboard/ts/action.ts
```diff
@@ -49,6 +50,25 @@ export const actionCreator = () => {
     setStudyDetailState(studyId, newStudy)
   }
 
+  const setTrialState = (
```
hvy, line 53: What do you think? To better describe its behavior.
```suggestion
  const setTrialStateValues = (
```

optuna_dashboard/ts/components/TrialTable.tsx
```diff
@@ -293,6 +337,56 @@ export const TrialTable: FC<{
             />
           </Box>
         </Grid>
+        {trials[index].state === "Running" ? (
+          <Grid item xs={12}>
+            <Box margin={1}>
+              <Typography variant="h6" gutterBottom component="div">
+                Trial tell operations
```
hvy, line 344: Cosmetic suggestion, I believe this communicates well enough and is more compact. What do you think?
```suggestion
                Trial tell
```

optuna_dashboard/ts/action.ts
```diff
@@ -265,6 +285,44 @@ export const actionCreator = () => {
...
+          (t) => t.trial_id === trialId
+        )
+        if (index === -1) {
+          enqueueSnackbar(`Unexpected error happens. Please reload the page.`, {
+            variant: "error",
+          })
+          return
+        }
+        setTrialState(studyId, index, state, values)
+        enqueueSnackbar(`Success to update trial (${message})`, {
```
hvy, line 310:
```suggestion
        enqueueSnackbar(`Successfully updated trial (${message})`, {
```

hvy (changes requested):
Mostly took a look at the server side and tried out the dashboard locally. Basically LGTM besides the small comments.

I noticed one thing, that I can enter values such as "1e" or "999A", that is, a number followed by a single char. It's not seemingly validated as an invalid input sequence, but nothing happens when you press "SUBMIT". Is this intended?

optuna_dashboard/_app.py
```diff
@@ -55,6 +56,13 @@
 BottleView = TypeVar("BottleView", bound=Callable[..., BottleViewReturn])
 
 logger = logging.getLogger(__name__)
+str_to_trial_state = {
+    "Running": TrialState.RUNNING,
+    "Waiting": TrialState.WAITING,
+    "Complete": TrialState.COMPLETE,
+    "Pruned": TrialState.PRUNED,
+    "Fail": TrialState.FAIL,
+}
```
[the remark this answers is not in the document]
c-bata: Sounds good!

c-bata (approved):
LGTM after applying @hvy's suggestions.

hvy:
Looks like the lint is failing. Otherwise, LGTM.

PR #456, opened 2023-05-10
Update Box in TrialListDetail scrollable

### LINKED ISSUE #448: Can Attribute and Parameter fields be made scrollable?
# Feature Request

Vertical and horizontal scrolling would be more convenient, is it possible to achieve this?
Details are written below.

## Vertical scroll

When trying to do a human-in-the-loop, if there are too many parameters, the user has to scroll down each time, which reduces usability.
Is it possible to make the Parameter  (and others) field default height up to, for example, 5 high and the rest scroll vertically?

image

## Horizontal scroll

For example, if there is a very long User Attribute, as shown below, the entire page will be very wide.
When using a slider with human-in-the-loop, the slider width is adjusted to the width of the page, which is inconvenient for the user, because if you try to put a large value in the slider, the user will have to scroll very far to the side.
Is it possible to set the width of the User Attribute (and others) field to a specific value and have the rest of the field values scroll horizontally?

image

---

I know that too many variables and long UserAttributes are not always desirable when performing optimization, but it is helpful to have this kind of functionality.

Thanks in advance,

hrntsm:

## Reference Issues/PRs

Fix #448 

The content mentioned in the above issue has been modified as shown in the following image.
In order to enable horizontal scrolling, maxWidth is set to 1000px for now, but the value will be modified if wider is better, etc.
The maximum height is set to 150px, but I just made sure that about 6 lines are displayed, and will modify this one if necessary.

If "alignItems" is set to "center", values above the center will not be displayed when vertical scrolling is enabled. Therefore, it is removed.
The result is that each value is not centered but slightly above the center.

Screenshot 2023-05-10 at 21 21 29

### DIFF
--- a/optuna_dashboard/ts/components/TrialList.tsx
+++ b/optuna_dashboard/ts/components/TrialList.tsx
@@ -226,10 +226,13 @@ const TrialListDetail: FC<{
               ? "rgba(255, 255, 255, 0.05)"
               : "rgba(0, 0, 0, 0.05)",
           width: "100%",
+          maxWidth: "1000px",
+          maxHeight: "150px",
+          overflow: "auto",
           p: theme.spacing(0.5, 1),
           borderRadius: theme.shape.borderRadius * 0.2,
           display: "flex",
-          alignItems: "center",
+          whiteSpace: "nowrap",
         }}
       >
         {value}


optuna_dashboard/ts/components/TrialList.tsx
```diff
@@ -226,10 +226,13 @@ const TrialListDetail: FC<{
               ? "rgba(255, 255, 255, 0.05)"
               : "rgba(0, 0, 0, 0.05)",
           width: "100%",
+          maxWidth: "1000px",
```
c-bata, line 229: How about removing this for wide display users?

```suggestion
```

c-bata (commented):
@hrntsm Thank you for your pull request! Looks almost good to me. I left one comment though.

optuna_dashboard/ts/components/TrialList.tsx
hrntsm: @c-bata Thanks comment!

Without a maximum width setting, the window will be as wide as the length of the string.
I'm trying to fix a problem with a Human-in-the-loop Slider widget where the width of the slider is lengthened to match the length of the string.(see Horizontal scrolling part of the issue)
So I would like to remove the maximum width in the section you commented on,  and add a maximum value to the slider width.
what do you think?

(Maybe below line?)
https://github.com/optuna/optuna-dashboard/blob/main/optuna_dashboard/ts/components/TrialFormWidgets.tsx#L284

hrntsm: The following modifications are intended
Screenshot 2023-05-16 at 20 13 23

c-bata: @hrntsm Thank you for your explanation. Sounds reasonable!

c-bata (approved):
Thanks for the update! LGTM.

PR #496, opened 2023-06-11
Migrate integration tests from pyppeteer to pytest-praywright

keisuke-umezawa:

## Reference Issues/PRs

NA

Rewrite visual_regression_test by pytest playwright

You can generate screenshots by `pytest --screenshot=on e2e_tests` and videos by `pytest --screenshot=video e2e_tests`.

### DIFF
--- a/e2e_tests/conftest.py
+++ b/e2e_tests/conftest.py
@@ -1,11 +1,8 @@
 import threading
-from typing import Generator
 from wsgiref.simple_server import make_server
 
 import optuna
 from optuna_dashboard import wsgi
-from playwright.sync_api import Browser
-from playwright.sync_api import sync_playwright
 import pytest
 
 
@@ -200,11 +197,3 @@ def server(
         thread.join()
 
     request.addfinalizer(stop_server)
-
-
-@pytest.fixture(scope="module")
-def browser() -> Generator[Browser, None, None]:
-    with sync_playwright() as playwright:
-        browser = playwright.chromium.launch()
-        yield browser
-        browser.close()
--- a/e2e_tests/conftest.py
+++ b/e2e_tests/conftest.py
@@ -1,11 +1,17 @@
-import pytest
-from playwright.sync_api import sync_playwright
-
 import threading
+from typing import Generator
 from wsgiref.simple_server import make_server
 
 import optuna
 from optuna_dashboard import wsgi
+from playwright.sync_api import Browser
+from playwright.sync_api import sync_playwright
+import pytest
+
+
+@pytest.fixture(scope="session")
+def port() -> int:
+    return 8081
 
 
 @pytest.fixture(scope="session")
@@ -180,18 +186,16 @@ def dummy_storage() -> optuna.storages.InMemoryStorage:
     return storage
 
 
-host = "127.0.0.1"
-port = 8081
-
-
-@pytest.fixture(scope='session')
-def server(request, dummy_storage):
+@pytest.fixture(scope="session")
+def server(
+    request: pytest.FixtureRequest, dummy_storage: optuna.storages.InMemoryStorage, port: int
+) -> None:
     app = wsgi(dummy_storage)
-    httpd = make_server(host, port, app)
+    httpd = make_server("127.0.0.1", port, app)
     thread = threading.Thread(target=httpd.serve_forever)
     thread.start()
 
-    def stop_server():
+    def stop_server() -> None:
         httpd.shutdown()
         httpd.server_close()
         thread.join()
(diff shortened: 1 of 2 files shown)


keisuke-umezawa:
@c-bata 
I fixed this PR. Could you check it?

e2e_tests/conftest.py
```diff
@@ -0,0 +1,210 @@
+import threading
```
keisuke-umezawa, line 1: ref: conftest.py
[Scope: sharing fixtures across classes, modules, packages or session](https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session)

e2e_tests/conftest.py
```diff
@@ -0,0 +1,202 @@
...
+    optuna.create_study(
+        study_name="multi-no-trials",
+        storage=storage,
+        directions=["minimize", "maximize"],
+        sampler=sampler,
+    )
+    return storage
+
+
+@pytest.fixture(scope="session")
```
keisuke-umezawa, line 188: See https://til.simonwillison.net/pytest/subprocess-server

e2e_tests/visual_regression_test.py
```diff
@@ -1,25 +1,29 @@
-import re
-
-from playwright.sync_api import expect
+import optuna
 from playwright.sync_api import Page
+import pytest
 
 
-def test_homepage_has_Playwright_in_title_and_get_started_link_linking_to_the_intro_page(
+@pytest.mark.parametrize("study_id", range(10))
```
keisuke-umezawa, line 6: It is difficult to get study_size from fixtures, and spcified 10 here as magic number.

e2e_tests/conftest.py
```diff
@@ -0,0 +1,199 @@
...
+from wsgiref.simple_server import make_server
+
+import optuna
+from optuna_dashboard import wsgi
+import pytest
+
+
+@pytest.fixture(scope="session")
+def port() -> int:
+    return 8081
```
c-bata, line 11: How about using more larger address number to avoid the port conflict.
 
```suggestion
    port = 38080
```

Or we can find a free port like below.
https://www.programcreek.com/python/?CodeExample=get+free+port

e2e_tests/visual_regression_test.py
```diff
@@ -1,25 +1,29 @@
...
+def test_study_list(
+    study_id: int,
     page: Page,
+    dummy_storage: optuna.storages.InMemoryStorage,
+    port: int,
+    server: None,
```
c-bata, line 12: Does it sound reasonable to return the base url like "http://localhost:{port}/dashboard/", then we can remove a `port` argument.

e2e_tests/visual_regression_test.py
```diff
@@ -1,25 +1,29 @@
...
 from playwright.sync_api import Page
+import pytest
 
 
-def test_homepage_has_Playwright_in_title_and_get_started_link_linking_to_the_intro_page(
+@pytest.mark.parametrize("study_id", range(10))
+def test_study_list(
+    study_id: int,
     page: Page,
+    dummy_storage: optuna.storages.InMemoryStorage,
```
c-bata, line 10: I have a question. Is `dummy_storage` created twice per each test since `server` fixture depends on `dummy_storage`.

c-bata (commented):
Thank you for your pull request. I left some suggestions.

e2e_tests/visual_regression_test.py
keisuke-umezawa: Fixed.

e2e_tests/conftest.py
keisuke-umezawa: Fixed.

e2e_tests/visual_regression_test.py
keisuke-umezawa: Actually, no if my understanding is correct.
The dummy storage and server fixture are created with `@pytest.fixture(scope="session")`. It means that they are created at the begging of all tests and will be destroyed at the end of the test session.

see more detail in https://docs.pytest.org/en/6.2.x/fixture.html#fixture-scopes

keisuke-umezawa:
@c-bata 
I followed your suggestions and answer your question. Could you check it again?

e2e_tests/conftest.py
```diff
@@ -0,0 +1,198 @@
...
+        sampler=sampler,
+    )
+    return storage
+
+
+@pytest.fixture(scope="session")
+def server_url(
+    request: pytest.FixtureRequest, dummy_storage: optuna.storages.InMemoryStorage
+) -> str:
+    ip = "127.0.0.1"
```
c-bata, line 184: A nit.

```suggestion
    addr = "127.0.0.1"
```

c-bata (approved):
LGTM with a nit.

PR #506, opened 2023-06-22
Fix check for single-objective study

contramundum53:
In `get_study_detail`, the check for single-objective study is buggy, and it performs `get_pareto_front_trials` even for single-objective study. Since `get_pareto_front_trials` is currently implemented as an O(#trials^2) operation, it becomes a bottleneck for study with >1000 trials.

This PR removes the bug and speeds up response time for single-objective study.


c-bata (approved):
LGTM!

PR #514, opened 2023-07-02
Add history graph use case in e2e tests

keisuke-umezawa:

### DIFF
--- a/e2e_tests/test_usecases/test_study_history.py
+++ b/e2e_tests/test_usecases/test_study_history.py
@@ -13,7 +13,7 @@ def test_history_xaxis_click(
     url = f"{server_url}/studies/{study_id}"
 
     page.goto(url)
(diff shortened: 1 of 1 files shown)


c-bata (approved):
LGTM.

However, I'd like to share some concerns. The current approach to write e2e_tests could lead to potential issues as the number of test cases increases. We need to create multiple studies within conftest.py for each test case although this test case only require the first study. So that it could substantially increases the time for running tests.

Moreover, I have a maintainability concern. If each test needs to assert more detailed conditions, it heavily depends on studies, set up in `conftest.py`. It makes tests complex. Ideally, it would be better to prepare only the minimum necessary studies for each test case.

keisuke-umezawa:
@c-bata
Thank you for the feedback! I will think about better solutions.

PR #517, opened 2023-07-05
Add slider to change the marker size

c-bata:

## Reference Issues/PRs

None

marker-size-change

### DIFF
--- a/optuna_dashboard/ts/components/GraphHistory.tsx
+++ b/optuna_dashboard/ts/components/GraphHistory.tsx
@@ -150,7 +150,7 @@ export const GraphHistory: FC<{
             defaultValue={5}
             marks={true}
             min={1}
-            max={20}
+            max={10}
             step={1}
             onChange={(e) => {
               // @ts-ignore
@@ -289,7 +289,7 @@ export const GraphHistoryMultiStudies: FC<{
             defaultValue={5}
             marks={true}
             min={1}
-            max={20}
+            max={10}
             step={1}
             onChange={(e) => {
               // @ts-ignore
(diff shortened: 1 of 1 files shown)


c-bata:
@HideakiImamura Could you review this PR?

HideakiImamura:
It seemed to me that we could focus on smaller areas. This is because it seems to me that we may want to reduce the size of a dot, but not to make it larger than it currently is.

c-bata:
Thanks for your input. I'm thinking of narrowing down the range to 1-10. What do you think?

HideakiImamura:
Sounds good. Could you apply the change?

optuna_dashboard/ts/components/GraphHistory.tsx
```diff
@@ -140,6 +144,20 @@ export const GraphHistory: FC<{
             />
           </RadioGroup>
         </FormControl>
+        <FormControl>
+          <FormLabel component="legend">Marker size:</FormLabel>
+          <Slider
+            defaultValue={5}
+            marks={true}
+            min={1}
+            max={20}
```
c-bata, line 153:
```suggestion
            max={10}
```

optuna_dashboard/ts/components/GraphHistory.tsx
```diff
@@ -263,6 +283,20 @@ export const GraphHistoryMultiStudies: FC<{
...
```
c-bata, line 292:
```suggestion
            max={10}
```

c-bata:
Applied. PTAL.

HideakiImamura (approved):
LGTM.

Issue #316, opened 2022-12-25
add constraints checkbox


okaikov:
# Feature Request

Hello,
I would appreciate if you can add the constraints functionality, and add a checkbox for choosing only trials that meet the constraints.
this functionality already exists in optuna.visualization.plot_pareto_front().
need to add this to: history & pareto front.

You can check the system attributes to see if it meets the constraints.

c-bata:
Sounds good! Thank you for your suggestion 👍

ldesmet:
Hello @okaikov @c-bata , my 2cents on this topic

1/ I don't think there is a checkbox in optuna.visualization.plot_pareto_front, only an argument to consider if trials satisfy constraints or not, but it plots all trials yet.

2/ This should be a new feature directly added to optuna right ? Like if we specify constraints in the sampler, it should consider these constraints in the pareto front directly ?

Have a great day,

okaikov:
AFAIK the actual constarints are stored at trial.SYSTEM_ATTR, no?

hrntsm:
Hi @okaikov 
I've implemented this in #525, please check it out!

c-bata:
@hrntsm Thank you for your pull request. Let me close this issue.

@okaikov Please feel free to reopen this issue if you have additional suggestions or feedbacks.

PR #525, opened 2023-07-21
Add constraints for pareto front

hrntsm:

## Reference Issues/PRs

Constraints have been supported for PareteFront in #316 

Since the constraint values are used in other plots, I use optuna's _CONSTRAINT_KEY to serialize the constraint values.

The results, reflecting the constraints, are shown below. A feasible trial has a gray line around the edge, while an infasible trial has a gray marker with no edge line.

Screenshot 2023-07-21 at 16 17 55

Screenshot 2023-07-21 at 16 24 39

### DIFF
--- a/optuna_dashboard/ts/apiClient.ts
+++ b/optuna_dashboard/ts/apiClient.ts
@@ -29,7 +29,8 @@ interface TrialResponse {
   user_attrs: Attribute[]
   system_attrs: Attribute[]
   note: Note
-  artifacts: Artifact[]
+  artifacts: Artifact[],
+  constraints: number[]
 }
 
 const convertTrialResponse = (res: TrialResponse): Trial => {
@@ -52,6 +53,7 @@ const convertTrialResponse = (res: TrialResponse): Trial => {
     system_attrs: res.system_attrs,
     note: res.note,
     artifacts: res.artifacts,
+    constraints: res.constraints,
   }
 }
 
--- a/optuna_dashboard/ts/apiClient.ts
+++ b/optuna_dashboard/ts/apiClient.ts
@@ -29,7 +29,7 @@ interface TrialResponse {
   user_attrs: Attribute[]
   system_attrs: Attribute[]
   note: Note
-  artifacts: Artifact[],
+  artifacts: Artifact[]
   constraints: number[]
 }
 
--- a/optuna_dashboard/ts/types/index.d.ts
+++ b/optuna_dashboard/ts/types/index.d.ts
@@ -112,6 +112,7 @@ type Trial = {
   }[]
   user_attrs: Attribute[]
   system_attrs: Attribute[]
+  constraints: number[]
   note: Note
   artifacts: Artifact[]
 }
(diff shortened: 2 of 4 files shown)


hrntsm:
In order to work with optuna v2.1, _CONSTRAINT_KEY, which was imported in _serializer.py, is no longer used and is defined as a constant in the file.

keisuke-umezawa:
@c-bata 
> In order to work with optuna v2.1, _CONSTRAINT_KEY, which was imported in _serializer.py, is no longer used and is defined as a constant in the file.

How long should we support optuna v2.1? 
If we do not need to support optuna v2.1 anymore, we can upgrade this version. https://github.com/optuna/optuna-dashboard/blob/main/.github/workflows/e2e-tests.yml#L19

optuna_dashboard/ts/apiClient.ts
```diff
@@ -30,6 +30,7 @@ interface TrialResponse {
   system_attrs: Attribute[]
   note: Note
   artifacts: Artifact[]
+  constraints: number[]
```
keisuke-umezawa, line 33: [nits] If we want to align with the other fields such as `values?`, should it be also an optional field?

optuna_dashboard/ts/types/index.d.ts
```diff
@@ -112,6 +112,7 @@ type Trial = {
   }[]
   user_attrs: Attribute[]
   system_attrs: Attribute[]
+  constraints: number[]
```
keisuke-umezawa, line 115: Ditto.

keisuke-umezawa (commented):
I left some minor comments and also request some comments to @c-bata . Overall, almost looks good to me!

hrntsm:
I have updated the value to an optional in response to your comment.

c-bata:
> @c-bata
> 
> > In order to work with optuna v2.1, _CONSTRAINT_KEY, which was imported in _serializer.py, is no longer used and is defined as a constant in the file.
> 
> How long should we support optuna v2.1? If we do not need to support optuna v2.1 anymore, we can upgrade this version. https://github.com/optuna/optuna-dashboard/blob/main/.github/workflows/e2e-tests.yml#L19

Thank you for asking. I think we can consider dropping Optuna v2 support from the next release 👍 Then, we need to replace [this line](https://github.com/optuna/optuna-dashboard/blob/8444a7058fde27b5b34d703fe81410c6f05f17c1/pyproject.toml#L30) with `"optuna>=3.0.0"` as well.

keisuke-umezawa:
@hrntsm 
Thank you for fixing it. But, I meant that as follows:

Currently, if there are not any constraints, the `constraints` become `[]`. But, to align the other fields such as `values?`, it might be better that the `constraints` become `undefined`.

However, I realized that it will introduce a lot of code changes, and there is not so much benefit to change `[]` to `undefined`. Sorry for disturbing you, but could you revert or remove [this commit](https://github.com/optuna/optuna-dashboard/pull/525/commits/7fbcb1900cd0e68bd5820babf653531039a0c55f)?

keisuke-umezawa (approved):

hrntsm:
I understand the intent of your comment. Thank you.
The commit itself has been reverted.

keisuke-umezawa:
@hrntsm 
Thank you for your quick fix! I merged it.

PR #552, opened 2023-08-12
Support 3dmodel artifacts

hrntsm:

This PR was intended to be an initial support for human-in-the-loop using 3d model generative AI and other methods.
If there are artifacts in the 3D model, the model can be displayed in 3d.
It is implemented using react-three-fiber, a library for react in threejs.
In order to make it easy to check the model, modal support is provided so that the model can be enlarged to the entire screen.

This PR supports only stl and 3dm file formats and normal mesh colors.
Therefore, it would be desirable to support other file formats and textures in the future.

Please refer to the following video for behavior.

https://github.com/optuna/optuna-dashboard/assets/23289252/2cd13b53-eb77-4aa8-adfc-26f8de9ae31b


optuna_dashboard/ts/components/ModelViewer.tsx
```diff
@@ -0,0 +1,109 @@
...
+const calculateBoundingBox = (geometries: THREE.BufferGeometry[]) => {
+  const boundingBox = new THREE.Box3()
+  geometries.forEach((geometry) => {
+    const mesh = new THREE.Mesh(geometry)
+    boundingBox.expandByObject(mesh)
+  })
+  return boundingBox
+}
+
+export function ModelViewer(props: ModelViewerProps): JSX.Element {
```
c-bata, line 37: I think the name `ModelViewer` is a bit ambiguous as it may be associated with machine learning "model".  How about naming it like below?

```suggestion
export const ThreejsArtifactViewer: React.FC<ModelViewerProps> = (props) => {
```

optuna_dashboard/ts/components/ModelViewer.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+interface ModelViewerProps {
+  src: string
+  width: string
+  height: string
+  hasGizmo: boolean
+  filetype: string | undefined
+}
+
+function CustomGizmoHelper(): JSX.Element {
```
c-bata, line 17: nit: In Optuna Dashboard, basically we use an arrow function syntax to declare React components.

```suggestion
const CustomGizmoHelper: React.FC<{}> = (props) => {
```

optuna_dashboard/ts/components/TrialList.tsx
```diff
@@ -437,6 +445,130 @@ const TrialArtifact: FC<{ trial: Trial }> = ({ trial }) => {
                 </CardContent>
               </Card>
             )
+          } else if (
+            modelViewerTargetExt.includes(a.filename.split(".").pop() || "")
```
c-bata, line 449: How about writing this condition like below to make the code more intuitive?

```suggestion
            a.filename.endswith(".stl") || a.filename.endswith(".3dm")
```

optuna_dashboard/ts/components/TrialList.tsx
```diff
@@ -437,6 +445,130 @@ const TrialArtifact: FC<{ trial: Trial }> = ({ trial }) => {
...
+                        return obj
+                      })
+                    }}
+                  >
+                    <FullscreenIcon />
+                  </IconButton>
+                  <Modal
+                    open={
+                      a.filename in open3dModelViewer
+                        ? open3dModelViewer[a.filename]
```
c-bata, line 514: We can upload multiple artifacts with the same filename. How about using `artifact_id` instead?

```suggestion
                        ? open3dModelViewer[a.artifact_id]
```

c-bata (commented):
Screenshot 2023-08-15 18 52 08

Awesome! Thank you for your pull request.
Changes looks almost good to me. I left some suggestions though.

optuna_dashboard/ts/components/TrialList.tsx
hrntsm: When increasing the number of supported formats in the future, I thought it would be easier to understand if they were all in one place, so I made a check (modelViewerTargetExt.includes) to see if they were included in the array.
However, since there are only a few more formats to be added, I'm going to modify the comment you made.

c-bata (approved):
Looks perfect to me.

PR #564, opened 2023-08-17
Deprecate artifact

Alnusjaponica:

* [ ] I agree this patch may be ported to [Goptuna](https://github.com/c-bata/goptuna) by other Goptuna contributors.

### DIFF
--- a/optuna_dashboard/artifact/_backend.py
+++ b/optuna_dashboard/artifact/_backend.py
@@ -66,7 +66,7 @@ def register_artifact_route(
         if artifact_store is None:
             response.status = 400  # Bad Request
             return b"Cannot access to the artifacts."
-        artifact_dict = get_artifact_meta(storage, study_id, trial_id, artifact_id)
+        artifact_dict = get_artifact_meta(storage, trial_id, artifact_id)
         if artifact_dict is None:
             response.status = 404
             return b"Not Found"
@@ -169,7 +169,6 @@ def upload_artifact(
     filename = os.path.basename(file_path)
     storage = trial.storage
     trial_id = trial._trial_id
-    study_id = trial.study._study_id
     artifact_id = str(uuid.uuid4())
     guess_mimetype, guess_encoding = mimetypes.guess_type(filename)
     artifact: ArtifactMeta = {
@@ -187,7 +186,7 @@ def upload_artifact(
 
 
 def get_artifact_meta(
-    storage: BaseStorage, study_id: int, trial_id: int, artifact_id: str
+    storage: BaseStorage, trial_id: int, artifact_id: str
 ) -> Optional[ArtifactMeta]:
     trial_system_attr = storage.get_trial_system_attrs(trial_id)
     attr_key = ARTIFACTS_ATTR_PREFIX + artifact_id
--- a/optuna_dashboard/artifact/_backend.py
+++ b/optuna_dashboard/artifact/_backend.py
@@ -102,8 +102,8 @@ def register_artifact_route(
             "mimetype": mimetype or DEFAULT_MIME_TYPE,
             "encoding": encoding,
         }
-        attr_key = _artifact_prefix(trial_id=trial_id) + artifact_id
-        storage.set_study_system_attr(study_id, attr_key, json.dumps(artifact))
+        attr_key = ARTIFACTS_ATTR_PREFIX + artifact_id
+        storage.set_trial_system_attr(trial_id, attr_key, json.dumps(artifact))
         response.status = 201
 
         trial = storage.get_trial(trial_id)
@@ -112,7 +112,7 @@ def register_artifact_route(
             return {"reason": "Invalid study_id or trial_id"}
         return {
             "artifact_id": artifact_id,
-            "artifacts": list_trial_artifacts(storage.get_study_system_attrs(study_id), trial),
+            "artifacts": list_trial_artifacts(storage.get_trial_system_attrs(trial_id), trial),
         }
 
     @app.delete("/api/artifacts/<study_id:int>/<trial_id:int>/<artifact_id:re:[0-9a-fA-F-]+>")
@@ -123,8 +123,8 @@ def register_artifact_route(
             return {"reason": "Cannot access to the artifacts."}
         artifact_store.remove(artifact_id)
 
-        attr_key = _artifact_prefix(trial_id) + artifact_id
-        storage.set_study_system_attr(study_id, attr_key, json.dumps(None))
+        attr_key = ARTIFACTS_ATTR_PREFIX + artifact_id
+        storage.set_trial_system_attr(trial_id, attr_key, json.dumps(None))
         response.status = 204
         return {}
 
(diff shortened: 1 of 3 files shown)


optuna_dashboard/artifact/_backend.py
```diff
@@ -178,32 +190,38 @@ def objective(trial: optuna.Trial) -> float:
...
     if artifact_meta is not None:
         return json.loads(artifact_meta)
 
+    # Search trial_system_attrs. Note that artifacts uploaded via optuna.artifacts.upload_artifact
+    # have a different trial_system_attrs key prefix.
     # See https://github.com/optuna/optuna/blob/f827582a8/optuna/artifacts/_upload.py#L71
     trial_system_attrs = storage.get_trial_system_attrs(trial_id)
-    value = trial_system_attrs.get("artifacts:" + artifact_id)
+    value = trial_system_attrs.get(
+        DASHBOARD_ARTIFACTS_ATTR_PREFIX + artifact_id
```
keisuke-umezawa, line 220: [question] Do we really need to `get(DASHBOARD_ARTIFACTS_ATTR_PREFIX + artifact_id)` because it seems that artifacts are uploaded with `ARTIFACTS_ATTR_PREFIX` from both optuna and optuna-dashboard?

https://github.com/optuna/optuna-dashboard/pull/564/files#diff-b37ec658ac0e92728ec1166294ddd30cf515cbb022e7d88fc970ab99bb8e6f82R113

optuna_dashboard/artifact/_backend.py
```diff
@@ -221,18 +239,20 @@ def delete_all_artifacts(backend: ArtifactStore, storage: BaseStorage, study_id:
...
     ]
 
+    # Collect ArtifactMeta from trial_system_attrs. Note that artifacts uploaded via
+    # optuna.artifacts.upload_artifacts have a different trial_system_attrs key prefix.
     # See https://github.com/optuna/optuna/blob/f827582a8/optuna/artifacts/_upload.py#L16
     optuna_artifact_metas = [
         json.loads(value)
         for key, value in trial.system_attrs.items()
-        if key.startswith("artifacts:")
+        if key.startswith(ARTIFACTS_ATTR_PREFIX) or key.startswith(DASHBOARD_ARTIFACTS_ATTR_PREFIX)
```
keisuke-umezawa, line 255: Ditto.

optuna_dashboard/artifact/_backend.py
Alnusjaponica: I noticed that it doesn't. Thanks for your comment.

optuna_dashboard/artifact/_backend.py
```diff
@@ -123,8 +127,17 @@ def delete_artifact(study_id: int, trial_id: int, artifact_id: str) -> dict[str,
             return {"reason": "Cannot access to the artifacts."}
         artifact_store.remove(artifact_id)
 
-        attr_key = _artifact_prefix(trial_id) + artifact_id
-        storage.set_study_system_attr(study_id, attr_key, json.dumps(None))
+        # The metadata of the artifact is stored in one of the following three locations:
+        storage.set_study_system_attr(
```
keisuke-umezawa, line 131: [question] Should we check that `_artifact_prefix(trial_id) + artifact_id` exists in study_system_attr?

optuna_dashboard/artifact/_backend.py
```diff
@@ -123,8 +127,17 @@ def delete_artifact(study_id: int, trial_id: int, artifact_id: str) -> dict[str,
...
+            study_id, _artifact_prefix(trial_id) + artifact_id, json.dumps(None)
+        )
+        storage.set_trial_system_attr(
+            trial_id, DASHBOARD_ARTIFACTS_ATTR_PREFIX + artifact_id, json.dumps(None)
+        )
+        storage.set_trial_system_attr(
```
keisuke-umezawa: [question] Should we check that `ARTIFACTS_ATTR_PREFIX + artifact_id` exists in set_trial_system_attr?

optuna_dashboard/artifact/_backend.py
```diff
@@ -123,8 +127,17 @@ def delete_artifact(study_id: int, trial_id: int, artifact_id: str) -> dict[str,
             return {"reason": "Cannot access to the artifacts."}
         artifact_store.remove(artifact_id)
 
-        attr_key = _artifact_prefix(trial_id) + artifact_id
-        storage.set_study_system_attr(study_id, attr_key, json.dumps(None))
+        # The metadata of the artifact is stored in one of the following three locations:
+        storage.set_study_system_attr(
+            study_id, _artifact_prefix(trial_id) + artifact_id, json.dumps(None)
+        )
+        storage.set_trial_system_attr(
```
keisuke-umezawa, line 134: ditto.

keisuke-umezawa (commented):
I left some comments. Could you check them?

optuna_dashboard/artifact/_backend.py
Alnusjaponica: I believe this is necessary to maintain backward compatibility since artifact meta information stored before this change are in  `study_system_attr`.

optuna_dashboard/artifact/_backend.py
Alnusjaponica: This is necessary since `artifact_store.remove()` does not remove artifact information from `trial_system_attr`.

optuna_dashboard/artifact/_backend.py
Alnusjaponica: This was already removed.

optuna_dashboard/artifact/_backend.py
keisuke-umezawa: Ok. I understand it.

It means that we will allow optuna-dashboard to be the following situations:

The behavior before this PR:
1. By using `upload_artifact_api`, key=`_artifact_prefix(trial_id) + artifact_id` in study_system_attr has some value.
2. By calling `delete_artifact`, key=`_artifact_prefix(trial_id) + artifact_id` in study_system_attr has `None`.

The behavior after this PR:
1. By using `upload_artifact_api`, key=`_artifact_prefix(trial_id) + artifact_id` in study_system_attr has some value with older version of optuna-dashboard, or  key=`ARTIFACTS_ATTR_PREFIX + artifact_id` in trial_system_attr has some value with newer version on optuna-dashboard.
2. By calling `delete_artifact`, both key=`_artifact_prefix(trial_id) + artifact_id` in study_system_attr and  key=`ARTIFACTS_ATTR_PREFIX + artifact_id` in trial_system_attr have  `None` even if they did not store any value in 1.

From my view, it is a little bit redundant that both study_system_attr and trial_system_attr have `None`, but it will work in the current implementation. I am ok with the current implementation.

keisuke-umezawa (approved):

python_tests/artifact/test_backend.py
```diff
@@ -0,0 +1,100 @@
...
+            MagicMock(
+                _trial_id=1, study=MagicMock(_study_id=study_id), system_attrs=trial1_system_attrs
+            ),
+        ]
+        if study_id == 0
+        else []
+    )
+    storage.get_all_trials.side_effect = get_all_trials
+
+    return storage
```
c-bata, line 62: Overusing `MagicMock` can decrease the readability and make tests more fragile, increasing maintenance costs. How about creating an actual `InMemoryStorage` object instead of using a Mock like the test case below?

https://github.com/optuna/optuna-dashboard/blob/f1259305b564884f1b14598185c2c569c639c0a3/python_tests/artifact/test_optuna_compatibility.py#L19-L53

c-bata (commented):
Thank you for your pull request. I left some comments.

c-bata (approved):
LGTM!

PR #639, opened 2023-09-30
Support for older versions of the schema

YuigaWada:

## Reference Issues/PRs

null

- Adds support for older versions of the schema. 
   - Especially, supported for  "v3.2.0.x", "v3.0.0.x" and "v2.6.0.x"
- Fixes bug regarding `intermediate_value_type`.

### DIFF
--- a/standalone_app/src/sqlite3.ts
+++ b/standalone_app/src/sqlite3.ts
@@ -63,14 +63,18 @@ const getSchemaVersion = (db: SQLite3DB): string => {
 }
 
 const isSupportedSchema = (schemaVersion: string): boolean => {
-  let lowestVersion = "v2.6.0.a"  // OK: "v3.2.0.a", "v3.0.0.d", "v2.6.0.a"
+  let lowestVersion = "v2.6.0.a" // supported: "v3.2.0.a", "v3.0.0.{a,b,c,d}", "v2.6.0.a"
   if (schemaVersion == lowestVersion) return true
-  return isGreaterSchemaVersion(schemaVersion,lowestVersion)
+  return isGreaterSchemaVersion(schemaVersion, lowestVersion)
 }
 
-const isGreaterSchemaVersion = (leftVersion: string, rightVersion: string): boolean => { // return leftVersion > rightVersion
-  leftVersion = leftVersion.replace(/\D/g,'')
-  rightVersion = rightVersion.replace(/\D/g,'')
+const isGreaterSchemaVersion = (
+  leftVersion: string,
+  rightVersion: string
+): boolean => {
+  // return leftVersion > rightVersion
+  leftVersion = leftVersion.replace(/\D/g, "")
+  rightVersion = rightVersion.replace(/\D/g, "")
   return Number(leftVersion) > Number(rightVersion)
 }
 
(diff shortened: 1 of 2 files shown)


standalone_app/src/sqlite3.ts
```diff
@@ -38,32 +38,47 @@ export const loadStorage = (
...
+
+const isGreaterSchemaVersion = (
+  leftVersion: string,
+  rightVersion: string
+): boolean => {
+  // return leftVersion > rightVersion
+  leftVersion = leftVersion.replace(/\D/g, "")
+  rightVersion = rightVersion.replace(/\D/g, "")
+  return Number(leftVersion) > Number(rightVersion)
 }
```
c-bata: Can we simply write this logic like below?

```suggestion
const isSupportedSchema = (schemaVersion: string): boolean => {
  return ["v3.2.0", "v3.0.0", "v2.6.0"].some((v) => schemaVersion.startsWith(v))
}
```

standalone_app/src/sqlite3.ts
```diff
@@ -233,15 +273,26 @@ const paramInternalValueToExternalValue = (
...
+    "UniformDistribution",
+    "LogUniformDistribution",
+    "DiscreteUniformDistribution",
+  ]
+  const intDistributionList = [
+    "IntDistribution",
+    "IntUniformDistribution",
+    "IntLogUniformDistribution",
+  ]
+  if (floatDistributionList.includes(parsed.name)) {
```
c-bata, line 293: The attributes of UniformDistribution, LogUniformDistribution, and DiscreteUniformDistribution, are bit different with FloatDistribution, so that we need to write this logic like below.

```python
if (parsed.name === "FloatDistribution") {
  return {
      type: "FloatDistribution",
      low: parsed.attributes.low as number,
      high: parsed.attributes.high as number,
      step: parsed.attributes.step as number,
      log: parsed.attributes.log as boolean,
    }
} else if (parsed.name === "UniformDistribution") {
  return {
      type: "FloatDistribution",
      low: parsed.attributes.low as number,
      high: parsed.attributes.high as number,
      step: null,
      log: false,
    }
} else if (parsed.name === "LogUniformDistribution") {
  return {
      type: "FloatDistribution",
      low: parsed.attributes.low as number,
      high: parsed.attributes.high as number,
      step: null,
      log: true,
    }
} else if (parsed.name === "DiscreteUniformDistribution") {
  return {
      type: "FloatDistribution",
      low: parsed.attributes.low as number,
      high: parsed.attributes.high as number,
      step: parsed.attributes.q,
      log: false,
    }
} else if ...
```

c-bata (commented):
Thank you for your pull request. I left two comments.

standalone_app/src/sqlite3.ts
YuigaWada: The `isGreaterSchemaVersion` function is used for switching to queries described later, so I think the number of lines will not significantly decrease further.

standalone_app/src/sqlite3.ts
YuigaWada: Understood. I'll rewrite the codes.

c-bata (approved):
LGTM!

PR #641, opened 2023-09-30
✨ feat: Use Vite instead of webpack

RE-yura:

Vite was installed in place of Webpack.
The server is started with Vite in development and served as a static file built with Vite in production by the API server.
Maintenance has been made a bit easier by cutting out the source code for the web client, which used to be in `optuna_dashboard/ts`, into `optuna_dashboard_client/web`.

- Unfinished tasks
  - Addition of README
  - Deletion of existing client-side files
    - In addition to `optuna_dashboard/ts`, it should be possible to remove package.json and other files in the root directory. Also, I would like to move the TypeScript related test files to a new directory.


optuna_dashboard/index.html
```diff
@@ -1,75 +1,123 @@
-<!DOCTYPE html>
+<!doctype html>
```
c-bata, line 1: Should we list this file on `.gitignore`?

c-bata:
Hello, thank you very much for your contribution, and I’m sorry that this PR has been left unattended for quite a while.

Since it has significantly diverged from the current main branch, it would require a substantial rework to merge. Unfortunately, I don’t have enough bandwidth to allocate to that effort right now. So I’m going to close this PR for now and open a new issue https://github.com/optuna/optuna-dashboard/issues/1230 to track the migration from Webpack to Vite.

The timeline for restarting this work is not decided yet, but when we resume, I’d be happy to include you as a co-author. Thank you again for your work and understanding 🙏

PR #649, opened 2023-10-06
Improve accuracy of `_one_side_trunc_norm_sampling`

not522:

Improve the accuracy of `_one_side_trunc_norm_sampling` and add tests. The expected values were calculated using `scipy.stats.truncnorm`.

### DIFF
--- a/optuna_dashboard/preferential/samplers/gp.py
+++ b/optuna_dashboard/preferential/samplers/gp.py
@@ -50,15 +50,15 @@ def _orthants_MVN_Gibbs_sampling(cov_inv: Tensor, cycles: int, initial_sample: T
 
 
 def _one_side_trunc_norm_sampling(lower: Tensor) -> Tensor:
-    if lower > 4.0:
-        r = torch.clamp_min(torch.rand(torch.Size(()), dtype=torch.float64), min=1e-300)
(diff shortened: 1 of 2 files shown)


not522:
I checked the accuracy using the following code.

```python
import matplotlib.pyplot as plt
import numpy as np
import scipy.special
import torch

def _one_side_trunc_norm_sampling(lower, r):
    r = torch.tensor([r], dtype=torch.float64)
    lower = torch.tensor([lower], dtype=torch.float64)
    return -torch.special.ndtri(torch.exp(torch.special.log_ndtr(-lower) + r.log()))

fig, ax= plt.subplots(3, 1)

rs = [0.5, 0.001, 0.999]
for i in range(3):
    r = rs[i]
    x = []
    y = []
    for lower in np.linspace(-10, 10, 100):
        x.append(lower)
        value = _one_side_trunc_norm_sampling(lower, r)
        expected = scipy.stats.truncnorm.ppf(1 - r, lower, float("inf"))
        y.append(float(value.numpy() - expected))
    ax[i].plot(x, y, label=f"r = {r}")
    ax[i].legend()

plt.subplots_adjust(hspace=0.6)
plt.show()
```

Figure_1

contramundum53 (approved):
LGTM!

PR #656, opened 2023-10-12
Add rank plot

### LINKED ISSUE #538: Implement rank plot
# Feature Request

The [rank plot](https://optuna.readthedocs.io/en/stable/reference/visualization/generated/optuna.visualization.plot_rank.html#optuna.visualization.plot_rank) was introduced in optuna v3.2.0 as an experimental feature. We would like to add this plot at the bottom of the [analytics page](https://github.com/optuna/optuna-dashboard/blob/main/optuna_dashboard/ts/components/StudyDetail.tsx#L86-L124) in the optuna dashboard. You can see the detailed implementation of the rank plot [here](https://github.com/optuna/optuna/blob/master/optuna/visualization/_rank.py).

HideakiImamura:

## Reference Issues/PRs

Resolve #538.

This PR implement the rank plot in the study analytics page. Since the rank plot is related to the relationship between parameters, it is placed between the contour plot and the edf plot. See below for example.
Screenshot 2023-10-12 at 15 48 38

The python script to create the study is as follows.
```python
import optuna


def objective(trial):
    x = trial.suggest_float("x", 0, 1)
    y = trial.suggest_float("y", 1e-2, 1e+0, log=True)
    z = trial.suggest_int("z", -2, 2)
    w = trial.suggest_float("w", 0, 1)

    return x + y + z + w


if __name__ == "__main__":
    study = optuna.create_study(storage="sqlite:///example.db")
    study.optimize(objective, n_trials=100)

```


c-bata:

contramundum53:
The color seems to be linear to the value, not rank.

HideakiImamura:
@contramundum53 Thanks for the comment. I fix the logic.

HideakiImamura:
@keisuke-umezawa Could you give me a review on this?

optuna_dashboard/ts/components/GraphRank.tsx
```diff
@@ -0,0 +1,342 @@
...
+    setYParam(param || null)
+  }
+
+  const rankPlotInfo = getRankPlotInfo(study, objectiveId, xParam, yParam)
+
+  useEffect(() => {
+    if (study != null) {
+      plotRank(rankPlotInfo, theme.palette.mode)
+    }
+  }, [study, theme.palette.mode])
```
keisuke-umezawa, line 73: To change ui in a timely manner, we need to change those lines something like this. Otherwise, when we change objectiveId, xParam and yParam, the ui will not be changed immediately.

```
  useEffect(() => {
    if (study != null) {
      const rankPlotInfo = getRankPlotInfo(study, objectiveId, xParam, yParam)
      plotRank(rankPlotInfo, theme.palette.mode)
    }
  }, [study, objectiveId, xParam, yParam, theme.palette.mode]
```

optuna_dashboard/ts/components/GraphRank.tsx
```diff
@@ -0,0 +1,342 @@
...
+  objectiveId: number,
+  xParam: SearchSpaceItem | null,
+  yParam: SearchSpaceItem | null
+): RankPlotInfo | null => {
+  if (study === null) {
+    return null
+  }
+
+  const trials = study.trials
+  const filtered_trials = trials.filter(filterFunc)
```
keisuke-umezawa, line 149: In this repo, we use lower camel case.

optuna_dashboard/ts/components/GraphRank.tsx
```diff
@@ -0,0 +1,342 @@
...
+  study: StudyDetail | null
+}> = ({ study = null }) => {
+  const theme = useTheme()
+  const [objectiveId, setobjectiveId] = useState<number>(0)
+  const searchSpace = useMergedUnionSearchSpace(study?.union_search_space)
+  const [xParam, setXParam] = useState<SearchSpaceItem | null>(null)
+  const [yParam, setYParam] = useState<SearchSpaceItem | null>(null)
+  const objectiveNames: string[] = study?.objective_names || []
+
+  if (xParam == null && searchSpace.length > 0) {
```
keisuke-umezawa, line 48: [question] Do you intentionally use `==` and `===` everywhere?

optuna_dashboard/ts/components/GraphRank.tsx
```diff
@@ -0,0 +1,342 @@
...
+  if (values.length > 2) {
+    color_idxs = raw_ranks.map((rank) => rank / (values.length - 1))
+  } else {
+    color_idxs = [0.5]
+  }
+  return color_idxs
+}
+
+const getOrderWithSameOrderAveraging = (values: number[]): number[] => {
+  const sorted_values = values.slice().sort()
```
keisuke-umezawa, line 273: Ditto. lowerCamelCase

optuna_dashboard/ts/components/GraphRank.tsx
```diff
@@ -0,0 +1,342 @@
...
+  const filtered_trials = trials.filter(filterFunc)
+  if (filtered_trials.length < 2 || xParam == null || yParam == null) {
+    return null
+  }
+
+  const xAxis = getAxisInfo(filtered_trials, xParam)
+  const yAxis = getAxisInfo(filtered_trials, yParam)
+
+  const xValues: number[] = []
+  const yValues: number[] = []
```
keisuke-umezawa, line 158: Do xValues and yValues support categorical values? In my local, the plot with categorical values seems not work well.

keisuke-umezawa (changes requested):
@HideakiImamura 
Sorry for being late. I just reviewed it. Could you check those comments?

optuna_dashboard/ts/components/GraphRank.tsx
HideakiImamura: Thank you for your comment. I don't have a specific policy, but I think we should standardize one way or the other. I will use `===`.

optuna_dashboard/ts/components/GraphRank.tsx
HideakiImamura: I fixed the type of the variables. Thanks for the comment.

I will check the behavior for the categorical variables.

keisuke-umezawa: Comparing the other plot, the rank plot has not supported categorical values yet.
スクリーンショット 2023-10-30 23 54 16
スクリーンショット 2023-10-30 23 54 23

HideakiImamura: I fixed the code to support the categorical variables. Basically, the order of the ticks is in dictionary order.

Screenshot 2023-10-31 at 14 46 04

If both the x-axis and y-axis are categorical variables, and there is no trial that yields the minimum values in dictionary order, the order of the ticks will not be in dictionary order. This is because we don't have a way to sort the ticks themselves. I believe we should accept this imperfection.

Screenshot 2023-10-31 at 14 49 58

HideakiImamura:
@keisuke-umezawa Thanks for your careful reviews. I fixed the code to support the categorical variables. PTAL.

keisuke-umezawa (approved):
LGTM! Thank you for working on it.

PR #663, opened 2023-10-19
Fix ThreejsArtifactViewer camera settings

hrntsm:

Fixed a camera setting that was wrong when displaying Threejs models. The fixes are as follows
- The camera now looks at the center of the model.
- The camera's drawing distance is set to an appropriate value.
- The camera position calculated from the model size is now reflected.

Because these settings were incorrect, models larger than 1 meter were not rendered correctly.


c-bata (approved):
Thank you for your pull request!

PR #678, opened 2023-11-02
Support constrained optimization for timeline plot

not522:

## Reference Issues/PRs

This feature corresponds to https://github.com/optuna/optuna/pull/4877.

This PR supports constrained optimization for timeline plot. Same as Optuna, it applies only to `COMPLETE` trials.

newplot (2)

### DIFF
--- a/optuna_dashboard/ts/components/GraphTimeline.tsx
+++ b/optuna_dashboard/ts/components/GraphTimeline.tsx
@@ -92,12 +92,7 @@ const plotTimeline = (trials: Trial[], mode: string) => {
     template: mode === "dark" ? plotlyDarkTemplate : {},
   }
 
-  const traces: Partial<plotly.PlotData>[] = []
-  for (const s of Object.keys(cm) as TrialState[]) {
-    const bars = trials.filter((t) => t.state === s)
-    if (bars.length === 0) {
-      continue
-    }
+  const makeTrace = (bars: Trial[], name: string, color: string) => {
     const starts = bars.map((b) => b.datetime_start ?? new Date())
     const completes = bars.map((b, i) => b.datetime_complete ?? starts[i])
     const trace: Partial<plotly.PlotData> = {
@@ -106,14 +101,38 @@ const plotTimeline = (trials: Trial[], mode: string) => {
       y: bars.map((b) => b.number),
       // @ts-ignore: To suppress ts(2322)
       base: starts.map((s) => s.toISOString()),
-      name: s,
+      name: name,
       text: bars.map((b) => makeHovertext(b)),
-      hovertemplate: "%{text}<extra>" + s + "</extra>",
+      hovertemplate: "%{text}<extra>" + name + "</extra>",
       orientation: "h",
-      marker: { color: cm[s] },
+      marker: { color: color },
       textposition: "none", // Avoid drawing hovertext in a bar.
     }
-    traces.push(trace)
+    return trace
+  }
+
+  const traces: Partial<plotly.PlotData>[] = []
+  for (const s of Object.keys(cm) as TrialState[]) {
+    const bars = trials.filter((t) => t.state === s)
+    if (bars.length === 0) {
+      continue
+    }
+    if (s == "Complete") {
+      const feasibleTrials = bars.filter((t) =>
+        t.constraints.every((c) => c <= 0)
+      )
+      const infeasibleTrials = bars.filter((t) =>
+        t.constraints.some((c) => c > 0)
+      )
+      if (feasibleTrials.length > 0) {
+        traces.push(makeTrace(feasibleTrials, "Complete", cm[s]))
+      }
+      if (infeasibleTrials.length > 0) {
+        traces.push(makeTrace(infeasibleTrials, "Infeasible", "#cccccc"))
+      }
+    } else {
+      traces.push(makeTrace(bars, s, cm[s]))
+    }
   }
   plotly.react(plotDomId, traces, layout)
 }
--- a/optuna_dashboard/ts/components/GraphTimeline.tsx
+++ b/optuna_dashboard/ts/components/GraphTimeline.tsx
@@ -112,12 +112,12 @@ const plotTimeline = (trials: Trial[], mode: string) => {
   }
 
   const traces: Partial<plotly.PlotData>[] = []
-  for (const s of Object.keys(cm) as TrialState[]) {
-    const bars = trials.filter((t) => t.state === s)
+  for (const [state, color] of Object.entries(cm)) {
+    const bars = trials.filter((t) => t.state === state)
     if (bars.length === 0) {
       continue
     }
-    if (s == "Complete") {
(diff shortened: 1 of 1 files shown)


not522:
@HideakiImamura @nabenabe0928 Could you review this PR?

optuna_dashboard/ts/components/GraphTimeline.tsx
```diff
@@ -106,14 +101,38 @@ const plotTimeline = (trials: Trial[], mode: string) => {
...
-      name: s,
+      name: name,
       text: bars.map((b) => makeHovertext(b)),
-      hovertemplate: "%{text}<extra>" + s + "</extra>",
+      hovertemplate: "%{text}<extra>" + name + "</extra>",
       orientation: "h",
-      marker: { color: cm[s] },
+      marker: { color: color },
       textposition: "none", // Avoid drawing hovertext in a bar.
     }
```
nabenabe0928, line 110: What about replacing `name` in the argument with `state`?

optuna_dashboard/ts/components/GraphTimeline.tsx
```diff
@@ -106,14 +101,38 @@ const plotTimeline = (trials: Trial[], mode: string) => {
...
+      )
+      if (feasibleTrials.length > 0) {
+        traces.push(makeTrace(feasibleTrials, "Complete", cm[s]))
+      }
+      if (infeasibleTrials.length > 0) {
+        traces.push(makeTrace(infeasibleTrials, "Infeasible", "#cccccc"))
+      }
+    } else {
+      traces.push(makeTrace(bars, s, cm[s]))
+    }
```
nabenabe0928, line 135: What about early return?

```suggestion
    if (s !== "Complete") {
      traces.push(makeTrace(bars, s, cm[s]))
      continue
    }
    const feasibleTrials = bars.filter((t) =>
      t.constraints.every((c) => c <= 0)
    )
    const infeasibleTrials = bars.filter((t) =>
      t.constraints.some((c) => c > 0)
    )
    if (feasibleTrials.length > 0) {
      traces.push(makeTrace(feasibleTrials, "Complete", cm[s]))
    }
    if (infeasibleTrials.length > 0) {
      traces.push(makeTrace(infeasibleTrials, "Infeasible", "#cccccc"))
    }
```

nabenabe0928 (commented):
Thank you for the PR!
I left some comments")

optuna_dashboard/ts/components/GraphTimeline.tsx
```diff
@@ -92,12 +92,7 @@ const plotTimeline = (trials: Trial[], mode: string) => {
     template: mode === "dark" ? plotlyDarkTemplate : {},
   }
 
-  const traces: Partial<plotly.PlotData>[] = []
-  for (const s of Object.keys(cm) as TrialState[]) {
-    const bars = trials.filter((t) => t.state === s)
-    if (bars.length === 0) {
-      continue
-    }
+  const makeTrace = (bars: Trial[], name: string, color: string) => {
```
nabenabe0928, line 95: It is a very minor comment, but what about replacing `name` with `state`?

```suggestion
  const makeTrace = (bars: Trial[], state: string, color: string) => {
```

optuna_dashboard/ts/components/GraphTimeline.tsx
HideakiImamura, line 135: This may cause a conflict with @nabenabe0928's comment, but how about the following to avoid the type-assertion (`a as T`? Type assertions may still leave a possibility of a runtime error, even if the code compiles successfully. This is due to the fact that type assertions do not reference any information at runtime and simply change the type of variables.

```suggestion
  for (const [state, color] of Object.entries(cm)) {
    const bars = trials.filter((t) => t.state === state)
    if (bars.length === 0) {
      continue
    }
    if (state == "Complete") {
      const feasibleTrials = bars.filter((t) =>
        t.constraints.every((c) => c <= 0)
      )
      const infeasibleTrials = bars.filter((t) =>
        t.constraints.some((c) => c > 0)
      )
      if (feasibleTrials.length > 0) {
        traces.push(makeTrace(feasibleTrials, "Complete", color))
      }
      if (infeasibleTrials.length > 0) {
        traces.push(makeTrace(infeasibleTrials, "Infeasible", "#cccccc"))
      }
    } else {
      traces.push(makeTrace(bars, state, color))
    }
  }
```

HideakiImamura (commented):
Thanks for the PR. I have a minor comment. PTAL.

HideakiImamura:
I confirmed that the change of this PR generate the following plot for the following script.

```python
    # Study with trials having certain duration and constriants
    def objective_with_duration(trial: optuna.Trial) -> float:
        trial.set_user_attr("constraint", [trial.number % 2])
    
        time.sleep(0.01)
        if trial.number % 3 == 0:
            raise optuna.TrialPruned()
        if trial.number % 7 == 0:
            return float("nan")
        return 0.0

    def constraints(trial: optuna.Trial) -> list[float]:
        return trial.user_attrs["constraint"]

    study = optuna.create_study(
        study_name="trials-with-duration", 
        storage=storage,
        sampler=optuna.samplers.NSGAIISampler(constraints_func=constraints),
    )

    trial = study.ask({"x": FloatDistribution(0, 10), "y": CategoricalDistribution(["Foo", "Bar"])})  # To create a running trial
    trial.set_user_attr("constraint", [1])
    time.sleep(0.01)

    study.optimize(objective_with_duration, n_trials=10)

    study.enqueue_trial({"x": 0, "y": 10})  # To create a waiting trial
    time.sleep(0.01)
```

Screenshot 2023-11-02 at 20 27 42

HideakiImamura:
I noticed that the running and waiting trials are not appropriately plotted. Also, it seems that the waiting trials are not plotted properly even in Optuna. As neither of these issues are due to the changes in this PR, I believe it would be best to address these in a separate PR.

| Optuna Dashboard | Optuna |
| --- | --- |
| Screenshot 2023-11-02 at 20 27 42 | timeline |

optuna_dashboard/ts/components/GraphTimeline.tsx
not522: I don't think early return is appropriate here since the codes inside the brackets are at the same level. For example, if we add another if statement for "Waiting", it should be like this.

```typescript
if (s === "Complete") {
  ...
} else if (s === "Waiting") {
  ...
} else {
  ...
}
```

not like this.

```typescript
if (s !== "Complete" && s !== "Waiting") {
  ...
  continue
}
if (s === "Complete") {
  ...
  continue
}
...
```

not522:
Thank you for your reviews. I have updated this PR responding to your comments. PTAL.

HideakiImamura (approved):
Thanks for the update. LGTM.

nabenabe0928 (approved):
Thanks for the update!  LGTM

keisuke-umezawa (approved):

PR #687, opened 2023-11-08
Fast pareto-front calculation for 2D

### LINKED ISSUE #63: Use log-linear algorithm to get pareto front trials.
See https://github.com/optuna/optuna/pull/2503 for details.

contramundum53:

## Reference Issues/PRs
Fixes #63.

### DIFF
--- a/optuna_dashboard/ts/components/GraphParetoFront.tsx
+++ b/optuna_dashboard/ts/components/GraphParetoFront.tsx
@@ -165,6 +165,60 @@ const makeMarker = (
   }
 }
 
+const getIsDominatedTrialND = (normalizedValues: number[][]) => {
+  // Fallback for straight-forward pareto front algorithm (O(N^2) complexity).
+  const dominatedTrials: boolean[] = []
+  normalizedValues.forEach((values0: number[], i: number) => {
+    const dominated = normalizedValues.some((values1: number[], j: number) => {
+      if (i === j) {
+        return false
+      }
+      return values0.every((value0: number, k: number) => {
+        return values1[k] <= value0
+      })
+    })
+    dominatedTrials.push(dominated)
+  })
+  return dominatedTrials
+}
+
+const getIsDominatedTrial2D = (normalizedValues: number[][]) => {
+  // Fast pareto front algorithm (O(N log N) complexity).
+  const sorted = normalizedValues
+    .map((values, i) => [values[0], values[1], i])
+    .sort()
+  let minValue1 = sorted[0][1]
+  const dominatedTrials: boolean[] = new Array(normalizedValues.length).fill(
+    true
+  )
+
+  sorted.forEach((values) => {
+    if (values[1] <= minValue1) {
+      dominatedTrials[values[2]] = false
+      minValue1 = values[1]
+    }
+  })
+  return dominatedTrials
+}
+
+const getIsDominatedTrial1D = (normalizedValues: number[][]) => {
+  const best_value = Math.min(...normalizedValues.map((values) => values[0]))
+  return normalizedValues.map((value) => value[0] !== best_value)
+}
+
+const getIsDominatedTrial = (normalizedValues: number[][]) => {
+  if (normalizedValues.length == 0) {
+    return []
+  }
+  if (normalizedValues[0].length == 1) {
+    return getIsDominatedTrial1D(normalizedValues)
+  } else if (normalizedValues[0].length == 2) {
+    return getIsDominatedTrial2D(normalizedValues)
+  } else {
+    return getIsDominatedTrialND(normalizedValues)
+  }
+}
+
 const plotParetoFront = (
   study: StudyDetail,
   objectiveXId: number,
@@ -218,18 +272,7 @@ const plotParetoFront = (
     }
   })
 
-  const dominatedTrials: boolean[] = []
-  normalizedValues.forEach((values0: number[], i: number) => {
-    const dominated = normalizedValues.some((values1: number[], j: number) => {
-      if (i === j) {
-        return false
-      }
-      return values0.every((value0: number, k: number) => {
-        return values1[k] <= value0
-      })
-    })
-    dominatedTrials.push(dominated)
-  })
+  const dominatedTrials: boolean[] = getIsDominatedTrial(normalizedValues)
 
   const plotData: Partial<plotly.PlotData>[] = [
     makeScatterObject(
--- a/optuna_dashboard/ts/components/GraphParetoFront.tsx
+++ b/optuna_dashboard/ts/components/GraphParetoFront.tsx
@@ -195,8 +195,8 @@ const getIsDominated2D = (normalizedValues: number[][]) => {
         ? -1
         : 0
     )
-  let maxValue0 = sorted[0][0]
-  let minValue1 = sorted[0][1]
+  let maxValueSeen0 = sorted[0][0]
+  let minValueSeen1 = sorted[0][1]
 
   const isDominated: boolean[] = new Array(normalizedValues.length).fill(false)
   sorted.forEach((values) => {
--- a/optuna_dashboard/ts/components/GraphParetoFront.tsx
+++ b/optuna_dashboard/ts/components/GraphParetoFront.tsx
@@ -181,20 +181,38 @@ const getIsDominatedTrialND = (normalizedValues: number[][]) => {
 }
 
 const getIsDominatedTrial2D = (normalizedValues: number[][]) => {
+  if (normalizedValues.length === 0) {
+    return []
+  }
   // Fast pareto front algorithm (O(N log N) complexity).
   const sorted = normalizedValues
     .map((values, i) => [values[0], values[1], i])
-    .sort()
+    .sort((a, b) =>
+      a[0] > b[0]
+        ? 1
+        : a[0] < b[0]
+        ? -1
+        : a[1] > b[1]
+        ? 1
+        : a[1] < b[1]
+        ? -1
+        : 0
+    )
+  let maxValue0 = sorted[0][0]
   let minValue1 = sorted[0][1]
-  const dominatedTrials: boolean[] = new Array(normalizedValues.length).fill(
-    true
-  )
 
+  const dominatedTrials: boolean[] = new Array(normalizedValues.length).fill(
+    false
+  )
   sorted.forEach((values) => {
-    if (values[1] <= minValue1) {
-      dominatedTrials[values[2]] = false
+    if (
+      values[1] > minValue1 ||
+      (values[1] >= minValue1 && values[0] > maxValue0)
+    ) {
+      dominatedTrials[values[2]] = true
       minValue1 = values[1]
     }
+    maxValue0 = values[0]
   })
   return dominatedTrials
 }
--- a/optuna_dashboard/ts/components/GraphParetoFront.tsx
+++ b/optuna_dashboard/ts/components/GraphParetoFront.tsx
@@ -208,7 +208,7 @@ const getIsDominatedTrial = (normalizedValues: number[][]) => {
   if (normalizedValues.length == 0) {
     return []
   }
-  if (normalizedValues[0].length == 1) {
+  if (normalizedValues[0].length === 1) {
     return getIsDominatedTrial1D(normalizedValues)
   } else if (normalizedValues[0].length == 2) {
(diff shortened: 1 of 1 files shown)


HideakiImamura:
@not522 @nabenabe0928 Could you review this PR?

nabenabe0928:
> @not522 @nabenabe0928 Could you review this PR?

Sure!

not522:
This is out of the scope of this PR, but the implementation for `dominatedTrials` is buggy when all values are same between two trials. They should be non-dominated, but it makes them dominated. Could you fix it?

not522 (commented):
I think the variable and function names sound unnatural. How about the followings?
`dominatedTrials` -> `dominationFlags`
`getIsDominatedTrial` -> `evaluateTrialDomination`

contramundum53:
I think renaming `dominatedTrials` to `isDominatedTrials` fixes all oddness about the naming. How about that?

not522:
I think starting a variable name with "is" may not be appropriate for an array. That is, "... is dominated trials" is grammatically incorrect.

optuna_dashboard/ts/components/GraphParetoFront.tsx
```diff
@@ -165,6 +165,58 @@ const makeMarker = (
...
+  )
+
+  sorted.forEach((values) => {
+    if (values[1] <= minValue1) {
+      dominatedTrials[values[2]] = false
+      minValue1 = values[1]
+    }
+  })
+  return dominatedTrials
+}
```
HideakiImamura, line 200: Is it OK when the following `sorted`?
```
sorted = [
    [0, 1, 0],
    [0, 2, 1],
    [0, 0, 2],
]
```

I think then `dominatedTrials = [false, true, false]`, but all trials are not dominated.

HideakiImamura (commented):
Thanks for the PR. I have several comments. PTAL.

optuna_dashboard/ts/components/GraphParetoFront.tsx
contramundum53: Thank you for pointing it out!
I'll change it as follows:

```suggestion
const getIsDominatedTrial2D = (normalizedValues: number[][]) => {
  // Fast pareto front algorithm (O(N log N) complexity).
  const sorted = normalizedValues
    .map((values, i) => [values[0], values[1], i])
    .sort()
  let minValue1 = sorted[0][1]
  const dominatedTrials: boolean[] = new Array(normalizedValues.length).fill(
    true
  )

  sorted.forEach((values) => {
    if (values[1] <= minValue1) {
      dominatedTrials[values[2]] = false
      minValue1 = values[1]
    }
  })
  return dominatedTrials
}
```
Then, `sorted` will be
```
sorted = [
    [0, 0, 2],
    [0, 1, 0],
    [0, 2, 1],
]
```
and [0, 0, 2] will be correctly flagged as undominated (other two are dominated).

contramundum53: @HideakiImamura How about this?

HideakiImamura (approved):
LGTM.

contramundum53:
> I think starting a variable name with "is" may not be appropriate for an array. That is, "... is dominated trials" is grammatically incorrect.
@not522 How about `isDominated` then?

not522:
> How about `isDominated` and `getIsDominated` then?

It sounds more natural. 👍

optuna_dashboard/ts/components/GraphParetoFront.tsx
```diff
@@ -165,6 +165,58 @@ const makeMarker = (
...
+  const sorted = normalizedValues
+    .map((values, i) => [values[0], values[1], i])
+    .sort()
+  let minValue1 = sorted[0][1]
+  const dominatedTrials: boolean[] = new Array(normalizedValues.length).fill(
+    true
+  )
+
+  sorted.forEach((values) => {
+    if (values[1] <= minValue1) {
```
not522, line 194: This condition is buggy when the values between trials are the same.

not522 (changes requested):

optuna_dashboard/ts/components/GraphParetoFront.tsx
contramundum53: In that case, both trials should be considered undominated.

optuna_dashboard/ts/components/GraphParetoFront.tsx
```diff
@@ -165,6 +165,58 @@ const makeMarker = (
...
+    dominatedTrials.push(dominated)
+  })
+  return dominatedTrials
+}
+
+const getIsDominatedTrial2D = (normalizedValues: number[][]) => {
+  // Fast pareto front algorithm (O(N log N) complexity).
+  const sorted = normalizedValues
+    .map((values, i) => [values[0], values[1], i])
+    .sort()
```
nabenabe0928, line 187:
```suggestion
    .sort((a, b) => a - b))
```

nabenabe0928 (changes requested):
Btw, it is still necessary.

optuna_dashboard/ts/components/GraphParetoFront.tsx
nabenabe0928, line 187:
```ts
const values = [[2, 2], [3, 2], [10, 2]]
const sortedValues = values.map((v, i) => [v[0], v[1], i]).sort()
console.log(sortedValues)
```

Output

```
[ [ 10, 2, 2 ], [ 2, 2, 0 ], [ 3, 2, 1 ] ]
```

So you still need to change the comparator inside `sort()`.

```suggestion
    .sort((a, b) => {
        if (a[0] !== b[0]){
            return a[0] < b[0]
        }else if (a[1] !== b[1]){
            return a[1] < b[1]
        }else{
            return a[2] < b[2]
        }
    })
```

nabenabe0928 (changes requested):

optuna_dashboard/ts/components/GraphParetoFront.tsx
nabenabe0928: Now given a sequence of 2D vectors $\\{(a_n, b_n)\\}_{n=1}^N$ such that $a_1 \leq a_2 \leq \dots \leq a_N$ and $b_i \leq b_j$ for all $\\{(i, j) \in [N] \times [N] \mid a_i = a_j, i < j\\}$ where $[N] \coloneqq \\{1,2,\dots,N\\}$, the condition here is a bit strange when the equality holds.

For example, $a_i \leq a_j$ holds from the assumption and we now assume that $b_i = b_j$.
As mentioned above, let's assume that $a_i = a_j$ and $b_i = b_j$ do not dominate each other.
However, $a_i < a_j$ and $\min_{n \in [i]} b_i = b_j$ could still hold.
It means that there exists $(a_i, b_i)$ such that $a_i < a_j$ and $b_i \leq b_j$, which weakly dominates $(a_j, b_j)$.

Maybe I am wrong, if so, correct me please!

nabenabe0928: Quick solution to this is the following:

```ts
const normalizedValues = [[0, 1], [1, 1], [2, 1]]

const getIsDominatedTrial2D = (normalizedValues: number[][]) => {
    const sortedValues = normalizedValues
      .map((values, i) => [values[0], values[1], i])
      .sort((a, b) => a[0] !== b[0] ? a[0] - b[0] : a[1] - b[1])
    let minValue1 = sortedValues[0][1]
    let latestParetoSolution = sortedValues[0].slice(0, 2)
    const dominatedTrials: boolean[] = new Array(normalizedValues.length).fill(
      true
    )
    sortedValues.forEach((values) => {
      if (values[1] < minValue1) {
        dominatedTrials[values[2]] = false
        minValue1 = values[1]
        latestParetoSolution = values.slice(0, 2)
      }else if (latestParetoSolution.every((v, i) => v === values[i])) {
        dominatedTrials[values[2]] = false
      }
    })
    return dominatedTrials
}

console.log(getIsDominatedTrial2D(normalizedValues))
```

optuna_dashboard/ts/components/GraphParetoFront.tsx
```diff
@@ -165,6 +165,58 @@ const makeMarker = (
...
+      dominatedTrials[values[2]] = false
+      minValue1 = values[1]
+    }
+  })
+  return dominatedTrials
+}
+
+const getIsDominatedTrial1D = (normalizedValues: number[][]) => {
+  const best_value = Math.min(...normalizedValues.map((values) => values[0]))
+  return normalizedValues.map((value) => value[0] !== best_value)
```
nabenabe0928, line 204:
```suggestion
  return normalizedValues.map((values) => values[0] !== best_value)
```

optuna_dashboard/ts/components/GraphParetoFront.tsx
HideakiImamura: See https://github.com/optuna/optuna/blob/master/optuna/study/_multi_objective.py#L101. Optuna uses `weakly dominates` for comparison. It seems that @nabenabe0928 is correct.

not522: Thank you for pointing out. My statement is incorrect and @nabenabe0928's one is correct, I think.

contramundum53:
@nabenabe0928 @not522 Thank you for pointing them out. I fixed the bug. PTAL.

optuna_dashboard/ts/components/GraphParetoFront.tsx
nabenabe0928: The new modification fails with the following case. 

```tsc
const normalizedValues = [[0, 1], [1, 2], [2, 1], [3, 0], [3, 0]]
```

We get:

```
[ false, true, false, false, false ]
```

But we would like to have:

```
[ false, true, true, false, false ]
```

optuna_dashboard/ts/components/GraphParetoFront.tsx
```diff
@@ -165,6 +165,71 @@ const makeMarker = (
...
+  const isDominated: boolean[] = new Array(normalizedValues.length).fill(false)
+  sorted.forEach((values) => {
+    if (
+      values[1] > minValue1 ||
+      (values[1] >= minValue1 && values[0] > maxValue0)
+    ) {
+      isDominated[values[2]] = true
+      minValue1 = values[1]
+    }
+    maxValue0 = values[0]
```
nabenabe0928, line 210: Or probably, it should work as well.
```suggestion
    }
    maxValue0 = values[0]
    minValue1 = Math.min(...[minValue1, values[1]])
```

optuna_dashboard/ts/components/GraphParetoFront.tsx
```diff
@@ -165,6 +165,71 @@ const makeMarker = (
...
+        : 0
+    )
+  let maxValue0 = sorted[0][0]
+  let minValue1 = sorted[0][1]
+
+  const isDominated: boolean[] = new Array(normalizedValues.length).fill(false)
+  sorted.forEach((values) => {
+    if (
+      values[1] > minValue1 ||
+      (values[1] >= minValue1 && values[0] > maxValue0)
```
nabenabe0928, line 205: I guess it is clearer as the inequality is always removed by the first condition.
```suggestion
      (values[1] === minValue1 && values[0] > maxValue0)
```

optuna_dashboard/ts/components/GraphParetoFront.tsx
```diff
@@ -165,6 +165,71 @@ const makeMarker = (
...
+        : a[0] < b[0]
+        ? -1
+        : a[1] > b[1]
+        ? 1
+        : a[1] < b[1]
+        ? -1
+        : 0
+    )
+  let maxValue0 = sorted[0][0]
+  let minValue1 = sorted[0][1]
```
nabenabe0928, line 199: This change is optional, if you think it is better, please take it, but otherwise, please just ignore it!
```suggestion
  let maxValueSeen0 = sorted[0][0]
  let minValueSeen1 = sorted[0][1]
```

nabenabe0928 (approved):
LGTM!

contramundum53:
@not522 Thank you for your review. I think I fixed the problems, PTAL.

not522 (approved):
Thank you for your update. I confirmed that this PR makes the Pareto-front calculation faster when many trials are best trials. LGTM.

PR #697, opened 2023-11-13
Fix audio artifact component

contramundum53:

The layout of audio artifact component was broken so I fixed it.

image


c-bata (approved):
Thank you for your pull request. I've noticed that there is a similar issue when uploading `.zip` files, but it should be handled in the following PR. LGTM!

Screenshot 2023-11-13 19 45 34

PR #703, opened 2023-11-18
Ensure trial notes are transferred when renamed

### LINKED ISSUE #660: Renaming studies doesn't preserve their trials' markdown notes
### Description

When renaming a study from the landing page of the dashboard by clicking on the rename button as follows:
image

the new study gets created and the trials are copied over, however, without each trials' notes, the notes get lost in the process.

### How to Reproduce

Add a note to any study's trial, then rename that study from the dashboard's landing page to observe that note to disappear subsequently.

Happy to provide more reproducibility info, in case it's an issue only on my side.

### Python version

3.9.16

### Optuna version

3.3.0

0.12.0

### Web browser

Firefox

adjeiv:

## Reference Issues/PRs
Fixes #660 

This PR defines a few more functions in _note, to delete and transfer notes for a study/trial. Tests are also added, and I've tested this manually locally.

https://github.com/optuna/optuna-dashboard/assets/52001888/e0a25d2e-1d72-4117-ab44-debf5c47251b


optuna_dashboard/_note.py
```diff
@@ -110,6 +110,23 @@ def note_str_key_prefix(trial_id: Optional[int]) -> str:
...
+        note = get_note_from_system_attrs(system_attrs, src_trial_id)["body"]
+        save_note_with_version(storage, dst_study._study_id, dst_trial_id, 0, note)
+        delete_notes(storage, src_study._study_id, src_trial_id)
+
+    # Transfer individual trial notes
+    for src_trial, dst_trial in zip(src_study.get_trials(), dst_study.get_trials()):
+        transfer(src_trial._trial_id, dst_trial._trial_id)
+
+    # Transfer study note
+    NO_SRC_TRIAL, NO_DST_TRIAL = None, None
```
c-bata, line 126: Could you simply transfer all `system_attrs` values like below?

```
system_attrs = storage.get_study_system_attrs(study_id=src_study._study_id)
for key, value in system_attrs.items():
    storage.set_study_system_attrs(study_id=dst_study._study_id)
```

c-bata (changes requested):
Thank you for your pull request. I left a comment.

adjeiv:
Thank you for the swift review. 
This was my initial approach; however, the keys used for the notes embed the trial id within them. So when we create the new study, the we also get entirely new trial IDs, and so we wouldn't be able to find their notes.

Let me know if there's another method to work around this!

optuna_dashboard/_note.py
```diff
@@ -110,6 +110,23 @@ def note_str_key_prefix(trial_id: Optional[int]) -> str:
...
+    transfer(NO_SRC_TRIAL, NO_DST_TRIAL)
```
c-bata, line 127: I would say we don't need to delete original notes. How about making it `copy` instead of `transfer` since we can reduce the number of storage API calls?

```suggestion
def copy_notes(storage: BaseStorage, src_study: optuna.Study, dst_study: optuna.Study) -> None:
    system_attrs = storage.get_study_system_attrs(study_id=src_study._study_id)

    # Copy individual trial notes
    for src_trial, dst_trial in zip(src_study.get_trials(), dst_study.get_trials()):
        note = get_note_from_system_attrs(system_attrs, src_trial._trial_id)["body"]
        save_note_with_version(storage, dst_study._study_id, dst_trial._trial_id, 0, note)

    # Copy study note
    note = get_note_from_system_attrs(system_attrs, None)["body"]
    save_note_with_version(storage, dst_study._study_id, None, 0, note)
```

optuna_dashboard/_app.py
```diff
@@ -176,6 +177,7 @@ def delete_study(study_id: int) -> dict[str, Any]:
             delete_all_artifacts(artifact_store, storage, study_id)
 
         try:
+            note.delete_study_notes(storage, study_id)
```
c-bata, line 180: `study_system_attrs` will be removed in `storage.delete_study()` so that you don't need to delete notes here.

```suggestion
```

c-bata (commented):
Understood, you're right. Maybe the notes should have used the trial.number instead of the trial_id as the key, but maintaining backward compatibility with that change would be difficult. So I think it would be best to merge this PR first.

I left some review comments.

optuna_dashboard/_app.py
```diff
@@ -165,6 +165,7 @@ def rename_study(study_id: int) -> dict[str, Any]:
             response.status = 500
             return {"reason": "Failed to load the new study"}
 
+        note.copy_notes(storage, src_study, dst_study)
```
c-bata, line 168: Could you move this line to under the L155 to make it show error messages to users?

c-bata (commented):
Thanks for the update! Looks great to me. I left one minor comment though.

c-bata (approved):
Thanks for the update! LGTM.

PR #717, opened 2023-11-29
_cli.py: apply expandvars() to `storage` argument

nodakai:

so that I can pass a database password as an environment variable, like:

```
$ export PG_PASS=53cr37  # or k8s secret or whatever
$ optuna-dashboard.py 'postgresql://optuna:$PG_PASS@db/optuna'
```

Note we must not let Bash interpret the `$` notation as that'll expose the secret to other users on the machine through commands like `ps`, `top`

Apologies, but I haven’t had the opportunity to test it myself...

### DIFF
--- a/optuna_dashboard/_cli.py
+++ b/optuna_dashboard/_cli.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import argparse
 import os
+import os.path
 from socketserver import ThreadingMixIn
 import sys
 from typing import TYPE_CHECKING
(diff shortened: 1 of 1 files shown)


c-bata:
@not522 Could you review this PR?

nodakai:
Hmm, just realized that the `optuna` CLI (but not APIs) supports `$OPTUNA_STORAGE` as an experimental feature, accepting a full SQLAlchemy URI.

I think my patch addresses a similar scenario, but it requires the user to explicitly specify `'$OPTUNA_STORAGE'` (with single quotes). In contrast, with the `optuna` CLI, it's possible to omit `--storage ...` when `$OPTUNA_STORAGE` is set.

nabenabe0928:
@nodakai If the solution suggested by @not522 works for you, we would like to close this PR.
Could you confirm this with a comment rather than with an emoji?

nodakai:
Yes, please go ahead with https://github.com/optuna/optuna-dashboard/pull/717#issuecomment-1835609946

PR #727, opened 2023-12-07
Clear cache before going to another tests in unit tests

nabenabe0928:

## Reference Issues/PRs

As the unit tests take over the cache from previous tests, I added a clearing function for such cache.
By adding this to the current set of the unit tests, we can test our code more safely.

I added a function for cleaning up the cache.

### DIFF
--- a/e2e_tests/test_dashboard/test_usecases/test_preferential_optimization.py
+++ b/e2e_tests/test_dashboard/test_usecases/test_preferential_optimization.py
@@ -9,6 +9,7 @@ from playwright.sync_api import Page
 import pytest
 
 from ...test_server import make_test_server
+from ...utils import clear_inmemory_cache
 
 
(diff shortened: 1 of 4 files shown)


c-bata:

nabenabe0928:
To reproduce the failure of this test, you can insert the following after [this line](https://github.com/optuna/optuna-dashboard/blob/main/optuna_dashboard/_storage.py#L29):
```python
        if trials is not None:
            trials_from_study = storage.get_all_trials(study_id=study_id)
            assert trials[0] == trials_from_study[0], f"{trials[0]}, {trials_from_study[0]}"
```

And then run `pytest -W ignore e2e_tests/test_dashboard/visual_regression_test.py::test_trial_note`.

Without the modification, we get the assertion error.

keisuke-umezawa (approved):
LGTM. As a followup, I will think the way to delete cache when the server stops.

PR #728, opened 2023-12-08
Replace the time heuristic for storage check with check by a new trial number

nabenabe0928:

## Reference Issues/PRs

Since we currently refer to the storage to get all trials for GET requests, the load for the GET requests is often heavy.
By design, it is surely efficient to fetch all the trials when we would like to get many trials at once while it is quicker to fetch a trial if we just would like to know the state or existence of a trial.

The unittests of this PR pass after merging [this PR](https://github.com/optuna/optuna-dashboard/pull/727).
The previous PR fixes bugs in unittests, which caused leaked cache from unrelated tests.

I use the trick explained above to judge whether we should fetch all the trials.
Currently, we rely on the last timing when we check all the trials, but I replace this heuristic with the following:
1. Check the maximum trial number in cache (and define it as `max_trial_number`),
2. Check whether the storage has a trial with the number of `max_trial_number + 1` for the current `study_id` of interest,
3. If yes, fetch all the trials from the storage as it is necessary,
4. If not, check all the cached trials' states, and if any states are updatable and the corresponding trial in the storage is not updatable (`COMPLETE`, `PRUNED`, or `FAILED`), we fetch all the trials,
5. Otherwise, we do not check all the trials.

Note that we need the time complexity of `O(N)` where `N` is the number rows in the storage to fetch a trial and I assume that we do not have so many running or waiting trials in each study as it is very unlikely to have many of such trials if Optuna works as intended.

### DIFF
--- a/e2e_tests/utils.py
+++ b/e2e_tests/utils.py
@@ -1,9 +1,7 @@
 from optuna_dashboard._storage import trials_cache
 from optuna_dashboard._storage import trials_cache_lock
-from optuna_dashboard._storage import trials_last_fetched_at
 
 
 def clear_inmemory_cache() -> None:
     with trials_cache_lock:
         trials_cache.clear()
-        trials_last_fetched_at.clear()
--- a/optuna_dashboard/_storage.py
+++ b/optuna_dashboard/_storage.py
@@ -40,6 +40,8 @@ def _should_update_trials_cache(storage: BaseStorage, study_id: int) -> bool:
         pass
 
     for t in trials:
+        # It is very likely that the number of updatable trials (running or waiting) is bounded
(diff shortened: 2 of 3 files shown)


c-bata:
Thank you for your huge effort :bow: Unfortunately I have some concerns regarding your proposed changes. 

Although the proposed logic improves the performance under certain scenarios, such as lower network latency or when studies have few running trials, it may issue numerous SQL queries when the study contains numerous unfinished trials that may lead to serious performance degradation. So, I would say the existing logic provides superior average performance with its simpler code.

I will be closing this pull request for now. If you have alternative ideas that could address these concerns, please feel free to reopen this. Once again, thank you for your time 🥇

PR #731, opened 2023-12-12
Make running trials in timeline plot visible

nabenabe0928:

## Reference Issues/PRs

This PR makes the running trials in the timeline plot visible.
Currently, running trials are handled as a run duration of zero.
For this reason, we cannot see running trials in the timeline plot although it actually exists.
In this PR, I aim to make them appear in the timeline plot.
Plus, I revised the code so that trials, which were killed before they completed, will handle the max time in the plot as the max completed/pruned/failed timing.

This PR makes the following changes:
1. Running trials will appear on the timeline plot, and
2. The maximum of the x-axis is handled as either the current date time (if the current running durations are not too long) or the maximum completed time.

### DIFF
--- a/optuna_dashboard/ts/components/GraphTimeline.tsx
+++ b/optuna_dashboard/ts/components/GraphTimeline.tsx
@@ -83,6 +83,10 @@ const plotTimeline = (trials: Trial[], mode: string) => {
       const start = t.datetime_start?.getTime() ?? new Date().getTime()
       const now = new Date().getTime()
       // This is an ad-hoc handling to check if the trial is running.
+      // We do not check via `trialState` because some trials may have state=RUNNING,
+      // even if they are not running because of unexpected job kills.
+      // In this case, we would like to ensure that these trials will not squash the timeline plot
+      // for the other trials.
       return now - start < maxRunDuration * 5
     })
   const maxDatetime = hasRunning
@@ -117,13 +121,13 @@ const plotTimeline = (trials: Trial[], mode: string) => {
   const makeTrace = (bars: Trial[], state: string, color: string) => {
     const isRunning = state === runningKey
     const starts = bars.map((b) => b.datetime_start ?? new Date())
-    const runDurations = bars.map((b) => {
-      const start = b.datetime_start?.getTime() ?? new Date().getTime()
-      const complete = b.datetime_complete?.getTime() ?? start
+    const runDurations = bars.map((b, i) => {
+      const startTime = starts[i].getTime()
+      const completeTime = b.datetime_complete?.getTime() ?? startTime
       // By using 1 as the min value, we can recognize these bars at least when zooming in.
       return Math.max(
         1,
-        !isRunning ? complete - start : maxDatetime.getTime() - start
+        !isRunning ? completeTime - startTime : maxDatetime.getTime() - startTime
       )
     })
     const trace: Partial<plotly.PlotData> = {
--- a/optuna_dashboard/ts/components/GraphTimeline.tsx
+++ b/optuna_dashboard/ts/components/GraphTimeline.tsx
@@ -65,10 +65,30 @@ const plotTimeline = (trials: Trial[], mode: string) => {
       )
     )
   )
-  const maxDatetime = new Date(
+  const maxRunDuration = Math.max(
+    ...lastTrials.map(
+      (t) => {
+        const start = t.datetime_start?.getTime() ?? new Date().getTime()
+        const complete = t.datetime_complete?.getTime() ?? start
+        return complete - start
+      }
+    )
+  )
+  const hasRunning = maxRunDuration === 0 || lastTrials.some((t) => {
+      const isRunning = t.state === "Running"
+      if (!isRunning){
+        return false
+      }
+      const start = t.datetime_start?.getTime() ?? new Date().getTime()
+      const now = new Date().getTime()
+      // This is an ad-hoc handling to check if the trial is running.
+      return now - start < maxRunDuration * 5
+    }
+  )
+  const maxDatetime = hasRunning ? new Date() : new Date(
     Math.max(
       ...lastTrials.map(
-        (t) => t.datetime_start?.getTime() ?? minDatetime.getTime()
+        (t) => t.datetime_complete?.getTime() ?? minDatetime.getTime()
       )
     )
   )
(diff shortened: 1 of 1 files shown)


c-bata:

nabenabe0928:
I checked the behavior with the following:

```python
import time

import optuna


def objective(trial: optuna.Trial) -> float:
    x = trial.suggest_float("x", -5, 5)
    time.sleep(30)
    return x


if __name__ == "__main__":
    study = optuna.create_study(storage="sqlite:///demo.db")
    study.optimize(objective, n_trials=100)

```

optuna_dashboard/ts/components/GraphTimeline.tsx
```diff
@@ -65,13 +66,34 @@ const plotTimeline = (trials: Trial[], mode: string) => {
...
+    (maxRunDuration === -Infinity &&
+      trials.some((t) => t.state === runningKey)) ||
+    trials.some((t) => {
+      if (t.state !== runningKey) {
+        return false
+      }
+      const start = t.datetime_start?.getTime() ?? new Date().getTime()
+      const now = new Date().getTime()
+      // This is an ad-hoc handling to check if the trial is running.
+      return now - start < maxRunDuration * 5
```
keisuke-umezawa, line 86: [question] Why do we need those lines for ad-hoc handling?

optuna_dashboard/ts/components/GraphTimeline.tsx
```diff
@@ -93,11 +115,20 @@ const plotTimeline = (trials: Trial[], mode: string) => {
   }
 
   const makeTrace = (bars: Trial[], state: string, color: string) => {
+    const isRunning = state === runningKey
     const starts = bars.map((b) => b.datetime_start ?? new Date())
-    const completes = bars.map((b, i) => b.datetime_complete ?? starts[i])
+    const runDurations = bars.map((b) => {
+      const start = b.datetime_start?.getTime() ?? new Date().getTime()
```
keisuke-umezawa, line 121: How about this one because we can always guarantee that `starts` and `start` are consistent?
```    
const runDurations = bars.map((b, i) => {
      const start = starts[i].getTime()
```

optuna_dashboard/ts/components/GraphTimeline.tsx
```diff
@@ -93,11 +115,20 @@ const plotTimeline = (trials: Trial[], mode: string) => {
...
+      const complete = b.datetime_complete?.getTime() ?? start
```
keisuke-umezawa, line 122: How about this since we can use the current time for running trials?
```
const complete = b.datetime_complete?.getTime() ?? maxDatetime.getTime()
```

optuna_dashboard/ts/components/GraphTimeline.tsx
```diff
@@ -93,11 +115,20 @@ const plotTimeline = (trials: Trial[], mode: string) => {
...
+      // By using 1 as the min value, we can recognize these bars at least when zooming in.
+      return Math.max(
+        1,
+        !isRunning ? complete - start : maxDatetime.getTime() - start
+      )
+    })
```
keisuke-umezawa, line 128: [question] Why do we need to get max here? If we fix the code as I said, those lines can be just `complete - start`?

keisuke-umezawa (approved):
I left come questions and comments. Could you check them?

keisuke-umezawa (changes requested):
Sorry, I wrongly added approved. I changed it to ng 🙇

optuna_dashboard/ts/components/GraphTimeline.tsx
keisuke-umezawa: For me, it can be just simply with `  const hasRunning = trials.some((t) => t.state === runningKey)`, but do you have some reasons that you need to write those lines?

optuna_dashboard/ts/components/GraphTimeline.tsx
keisuke-umezawa: For me, it seems that we can use the following lines:

```
    const runDurations = bars.map((b, i) => {
      const start = starts[i].getTime()
      const complete = b.datetime_complete?.getTime() ?? maxDatetime.getTime()
      return complete - start
    }
```

optuna_dashboard/ts/components/GraphTimeline.tsx
nabenabe0928: We can remove it as well, but then the problem is that the timeline for `RUNNING` trials, which was not somehow killed properly (it is basically preemption, which often happens, in our internal cluster) then the timeline plot for the other trials will not be visible because they will be squashed by the running trials, which are not really running because of preemption.

optuna_dashboard/ts/components/GraphTimeline.tsx
nabenabe0928: The lower bound `1` is necessary to be able to recognize at least bars exist in case `runDurations=0`.
Otherwise, I think your code is better.

nabenabe0928: Oh, actually, `const complete = b.datetime_complete?.getTime() ?? maxDatetime.getTime()` is incorrect when we think of `trialState===WAITING`.

optuna_dashboard/ts/components/GraphTimeline.tsx
nabenabe0928: `start` is necessary for waiting trials.

optuna_dashboard/ts/components/GraphTimeline.tsx
```diff
@@ -65,13 +66,38 @@ const plotTimeline = (trials: Trial[], mode: string) => {
...
   )
+  const hasRunning =
+    (maxRunDuration === -Infinity &&
+      trials.some((t) => t.state === runningKey)) ||
+    trials.some((t) => {
+      if (t.state !== runningKey) {
+        return false
+      }
+      const start = t.datetime_start?.getTime() ?? new Date().getTime()
+      const now = new Date().getTime()
```
keisuke-umezawa, line 84: [nits] We should define `now` on the top of this function, and use it instead of `Date()` everywhere to have consistent current time.

optuna_dashboard/ts/components/GraphTimeline.tsx
keisuke-umezawa: Ok, I understand that you want to show `waiting` and  in some cases also `running` trials.

In the optuna, it uses the current time for complete if a trial does not have it when it is `waiting` or `running`.
ref: https://github.com/optuna/optuna/blob/master/optuna/visualization/_timeline.py#L85-L86 Are you planning to make back port to optuna itself?

keisuke-umezawa (approved):
Thank you for working on it!
I just add a nits comment. If you fix it, you can merge it.
Btw, it makes some differences with optuna itself, and it is recommended to make back port to optuna.

optuna_dashboard/ts/components/GraphTimeline.tsx
nabenabe0928: Done

optuna_dashboard/ts/components/GraphTimeline.tsx
nabenabe0928: I did the backport [here](https://github.com/optuna/optuna/pull/5168)

nabenabe0928:
@keisuke-umezawa I made a backport to Optuna! Are there any other things I should do?

keisuke-umezawa:
@nabenabe0928 
The implementation in optuna side is not fixed yet, and it will make some differences between optuna and optuna-dashboad later. But, I believe that you will make a back port again from optuna to optuna-dashboard. So, I will merge this PR.

PR #732, opened 2023-12-15
Add journal storage loader

gen740:

## Reference Issues/PRs

None

- Add `journalStorage.ts` into standalone app, enabling `journal storage` support.

### DIFF
--- a/standalone_app/src/components/StorageLoader.tsx
+++ b/standalone_app/src/components/StorageLoader.tsx
@@ -30,12 +30,13 @@ export const StorageLoader: FC = () => {
     const r = new FileReader()
     r.addEventListener("load", () => {
       const arrayBuffer = r.result as ArrayBuffer | null
-      const fileExtension = file.name.split(".").pop()
       if (arrayBuffer !== null) {
-        if (fileExtension === "log") {
-          loadJournalStorage(arrayBuffer, setStudies)
-        } else {
+        const header = new Uint8Array(arrayBuffer, 0, 16)
+        const headerString = new TextDecoder().decode(header)
+        if (headerString === "SQLite format 3\u0000") {
           loadSQLite3Storage(arrayBuffer, setStudies)
+        } else {
+          loadJournalStorage(arrayBuffer, setStudies)
         }
       }
     })
--- a/standalone_app/src/components/StorageLoader.tsx
+++ b/standalone_app/src/components/StorageLoader.tsx
@@ -32,7 +32,7 @@ export const StorageLoader: FC = () => {
       const arrayBuffer = r.result as ArrayBuffer | null
       const fileExtension = file.name.split(".").pop()
       if (arrayBuffer !== null) {
-        if (fileExtension == "log") {
+        if (fileExtension === "log") {
           loadJournalStorage(arrayBuffer, setStudies)
         } else {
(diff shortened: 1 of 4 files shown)


nabenabe0928:
@c-bata Could you review this PR?

nabenabe0928:
@c-bata I removed you from the assignee of this PR.

standalone_app/src/components/StorageLoader.tsx
```diff
@@ -30,7 +31,13 @@ export const StorageLoader: FC = () => {
     r.addEventListener("load", () => {
       const arrayBuffer = r.result as ArrayBuffer | null
       if (arrayBuffer !== null) {
-        loadStorage(arrayBuffer, setStudies)
+        const header = new Uint8Array(arrayBuffer, 0, 16)
+        const headerString = new TextDecoder().decode(header)
+        if (headerString === "SQLite format 3\u0000") {
+          loadSQLite3Storage(arrayBuffer, setStudies)
+        } else {
+          loadJournalStorage(arrayBuffer, setStudies)
```
keisuke-umezawa, line 39: [question] Do you have any way to check that the file is compatible with JournalStorage? If yes, we can add "else if" to check it.

standalone_app/src/vscode_entry.tsx
```diff
@@ -35,7 +38,13 @@ export const AppWrapper: FC = () => {
             bytes[i] = binaryString.charCodeAt(i)
           }
           arrayBuffer = bytes.buffer
-          loadStorage(arrayBuffer, onceSetStudies)
+          header = new Uint8Array(arrayBuffer, 0, 16)
+          headerString = new TextDecoder().decode(header)
+          if (headerString === "SQLite format 3\u0000") {
+            loadSQLite3Storage(arrayBuffer, onceSetStudies)
+          } else {
+            loadJournalStorage(arrayBuffer, onceSetStudies)
```
keisuke-umezawa, line 46: ditto.

keisuke-umezawa (commented):
This is early review comment. I checked PR other than JournalStorage implementation. I will check it later 🙏 
Btw, I hope we can add some tests for it. At least, can you add e2e test for this one similar to this test?

https://github.com/optuna/optuna-dashboard/blob/38cc928df3423c08eb5dea60bd67d097d0e833b0/e2e_tests/test_standalone/test_study_list.py#L28-L74

standalone_app/src/components/StorageLoader.tsx
gen740: Unfortunately, there is no "header" or special "flag" in the JournalStorage log. The only requirement for JournalStorage is a JSON line file. We can detect the format by parsing the first line of the log, but I believe this should be handled in the `JournalStorage` class.

gen740:
Thank you for your comments!
Currently, Optuna-dashboard's standalone app lacks a test system. Consequently, there is no way to implement tests at this time. However, I plan to develop and add a testing system for the standalone app as a follow-up.

keisuke-umezawa:
@gen740 
Ok. In that case, I will quickly check this PR weather it will not break the entire system. If not, I will merge it soon.

keisuke-umezawa:
I confirmed that it also works well in my env. I merge it. Let's add some tests later.

PR #741, opened 2023-12-28
Fix the vscode extension's entry point for code-server execution

HideakiImamura:

## Motivations

The current VS Code extension does not support code-server. This PR aims to fix it.

## Details of the change

According to the [documentation](https://code.visualstudio.com/api/advanced-topics/remote-extensions#workarounds-for-using-localhost-from-a-webview), the VS Code extension for the code-server must not be the web extension. The current entry point is defined by `browser` property, which is for the web extension. I changed the property name from the `browser` to the `main`, which is for the standard extension.

## How to confirm the change

1. Create the VSIX file by running `make vscode-extension`.
2. Upload the VSIX file to the remote environment running the code-server.
3. Install the extension from the VSIX file. Right click the file, and press the `Install Extension VSIX`.
4. Open some sqlite file by right clicking and press the `Open in Optuna Dashboard`.

### DIFF
--- a/vscode/package-lock.json
+++ b/vscode/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "optuna-dashboard",
-  "version": "0.0.1",
+  "version": "0.1.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "optuna-dashboard",
-      "version": "0.0.1",
+      "version": "0.1.0",
       "license": "MIT",
       "devDependencies": {
         "@types/mocha": "^10.0.1",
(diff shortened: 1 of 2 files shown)


c-bata:
@HideakiImamura Thank you for your pull request.
I confirmed that this PR makes it being able to open the sqlite file. However, no studies are shown in my environment although I created some studies in the file. Is this problem reproducible in your environment?

Screenshot 2024-01-11 19 17 37

c-bata (approved):
I'm not sure why but now I can see that the extension works perfect. Thank you for your pull request!

Screenshot 2024-01-12 10 09 05

PR #742, opened 2023-12-30
Fix journal storage handling in windows

hrntsm:

## Reference Issues/PRs

https://github.com/optuna/optuna/pull/4308

In windows environment, if lock_obj is not specified explicitly, the same error as reference will occur when handling JournalStorage, so os is determined and lock_obj is included in the argument.

This is probably why the human-in-the-loop optimization does not work because the change to the journal storage  cannot be performed.


c-bata (approved):
LGTM.

PR #747, opened 2024-01-08
Change requirements to optuna>=3.1.0

keisuke-umezawa:
## Contributor License Agreement

This repository (``optuna-dashboard``) and [Goptuna](https://github.com/c-bata/goptuna) share common code.
This pull request may therefore be ported to Goptuna.
Make sure that you understand the consequences concerning licenses and check the box below if you accept the term before creating this pull request.

* [x] I agree this patch may be ported to [Goptuna](https://github.com/c-bata/goptuna) by other Goptuna contributors.

## Reference Issues/PRs

NA

## What does this implement/fix? Explain your changes.

Change requirements to optuna>=3.1.0

PR #723, opened 2023-12-06
Code fix/download csv

eukaryo:
## Contributor License Agreement

This repository (``optuna-dashboard``) and [Goptuna](https://github.com/c-bata/goptuna) share common code.
This pull request may therefore be ported to Goptuna.
Make sure that you understand the consequences concerning licenses and check the box below if you accept the term before creating this pull request.

* [x] I agree this patch may be ported to [Goptuna](https://github.com/c-bata/goptuna) by other Goptuna contributors.

## Reference Issues/PRs

I will take over this PR:
https://github.com/optuna/optuna-dashboard/pull/648

## What does this implement/fix? Explain your changes.

- added a function to make a csv file in backend.
- added a button to download the file in frontend. (in the bottom of TrialTable page)

PR #771, opened 2024-01-24
Skip unit test with optuna version

keisuke-umezawa:

## Reference Issues/PRs

To follow up on https://github.com/optuna/optuna-dashboard/pull/747

- I made unit tests pass with optuna v3.1.0.
- I also checked that unit test passed with v3.2.0 - latest in my local env.
- Added ci with oldest compatible optuna version v3.1.0

### DIFF
--- a/python_tests/artifact/test_backend.py
+++ b/python_tests/artifact/test_backend.py
@@ -128,6 +128,10 @@ def test_study_artifact_not_found() -> None:
         assert status == 404
 
 
+@pytest.mark.skipif(
+    version.parse(optuna.__version__) < version.parse("3.4.0"),
+    reason="upload_artiract needs storage",
+)
 def test_successful_study_artifact_retrieval() -> None:
     storage = optuna.storages.InMemoryStorage()
     study = optuna.create_study(storage=storage)
@@ -263,6 +267,10 @@ def test_upload_artifact() -> None:
             assert data == "dummy_content"
 
 
+@pytest.mark.skipif(
+    version.parse(optuna.__version__) < version.parse("3.4.0"),
+    reason="upload_artiract needs storage",
+)
 def test_delete_study_artifact() -> None:
     storage = optuna.storages.InMemoryStorage()
     study = optuna.create_study(storage=storage)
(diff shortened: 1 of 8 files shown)


python_tests/artifact/test_backend.py
```diff
@@ -122,6 +128,10 @@ def test_study_artifact_not_found() -> None:
         assert status == 404
 
 
+@pytest.mark.skipif(
+    version.parse(optuna.__version__) < version.parse("3.4.0"),
+    reason="upload_artiract needs storage",
+)
```
c-bata, line 134: In my understanding, this decorator will be ignored due to the module level `pytest.skip` call in L12. Is that correct?

python_tests/test_csv_download.py
```diff
@@ -5,11 +5,16 @@
...
 from optuna.trial import TrialState
 from optuna_dashboard._app import create_app
+from packaging import version
 import pytest
 
 from .wsgi_client import send_request
 
 
+if version.parse(optuna.__version__) < version.parse("3.4.0"):
+    pytest.skip("Study.metrics_name is introduced at v3.4.0", allow_module_level=True)
```
c-bata, line 15: [fyi] This dependency should be fixed, but it looks out of the scope of this PR.

https://github.com/optuna/optuna-dashboard/pull/723/files#diff-0f7b622fa29613e91a846a5635ff2339348136638848391f330189b81b224a80R471-R472

python_tests/artifact/test_backend.py
keisuke-umezawa:
```
if version.parse(optuna.__version__) < version.parse("3.3.0"):
    pytest.skip("optuna.artiracts module is introduced at v3.3.0", allow_module_level=True)
```

At the module level, we will skip tests < v3.3.0. But, only this test does not pass at v3.3.0. So, I added skipif only for this test.

c-bata: Ah, I understand. Thank you for your reply :pray:

c-bata (approved):
LGTM!

Issue #799, opened 2024-02-07
VSCode Extension doesn't show studies / Optuna Dashboard fails with `UnicodeDecodeError`


turbotimon:
### Description

I created a study as described [here](https://optuna.readthedocs.io/en/stable/tutorial/20_recipes/001_rdb.html#new-study) but the VSCode extension doesn't show it. If i open the db with a db viewer, the study is there:

![image](https://github.com/optuna/optuna-dashboard/assets/57718207/3c558ed6-fbf1-483d-a872-6235a02c8723)

Also `optuna-dashboard example-study.db` fails with an `UnicodeDecodeError`error (meight by related?):

```bash
$ optuna-dashboard nbs/example-study.db 
/home/user/projects/venv/lib/python3.11/site-packages/optuna_dashboard/_storage_url.py:88: ExperimentalWarning: JournalStorage is experimental (supported from v3.1.0). The interface can change in the future.
  return JournalStorage(JournalFileStorage(file_path=file_path))
Traceback (most recent call last):
  File "/home/user/projects/venv/bin/optuna-dashboard", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/user/projects/venv/lib/python3.11/site-packages/optuna_dashboard/_cli.py", line 119, in main
    storage = get_storage(args.storage, storage_class=args.storage_class)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/projects/venv/lib/python3.11/site-packages/optuna_dashboard/_storage_url.py", line 58, in get_storage
    return guess_storage_from_url(storage)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/projects/venv/lib/python3.11/site-packages/optuna_dashboard/_storage_url.py", line 66, in guess_storage_from_url
    return get_journal_file_storage(storage_url)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/projects/venv/lib/python3.11/site-packages/optuna_dashboard/_storage_url.py", line 88, in get_journal_file_storage
    return JournalStorage(JournalFileStorage(file_path=file_path))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #800, opened 2024-02-08
improve ux when trying to open sqlite without protocol

turbotimon:

## Reference Issues/PRs

Fixes/Improves UX for #799 

- Improves UX by guessing file using file header to open a sqlite db without protocol, e.g.:
```bash
$ optuna-dashboard example.db
```
- flake8 and mypy are happy with my changes
- i tried to add tests, but was not able because `pytest python_test/` fails on main with import errors

### DIFF
--- a/optuna_dashboard/_storage_url.py
+++ b/optuna_dashboard/_storage_url.py
@@ -59,10 +59,12 @@ def get_storage(
     return guess_storage_from_url(storage)
 
 
-def _has_sqlite_suffix(storage_url: str) -> bool:
+def _has_sqlite_header(storage_url: str) -> bool:
     storage_path = Path(storage_url)
-    SQLITE_EXTENSIONS = (".db", ".sqlite3")
-    return storage_path.suffix in SQLITE_EXTENSIONS
+    SQLITE_HEADER = b"SQLite format 3\x00" # see https://github.com/optuna/optuna-dashboard/pull/800
+    with storage_path.open(mode="rb") as f:
+        header = f.read(len(SQLITE_HEADER))
+    return header == SQLITE_HEADER
 
 
 def guess_storage_from_url(storage_url: str) -> BaseStorage:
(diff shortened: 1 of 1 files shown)


optuna_dashboard/_storage_url.py
```diff
@@ -58,11 +59,22 @@ def get_storage(
     return guess_storage_from_url(storage)
 
 
+def _has_sqlite_suffix(storage_url: str) -> bool:
+    storage_path = Path(storage_url)
+    SQLITE_EXTENSIONS = (".db", ".sqlite3")
```
c-bata, line 64: Actually, Optuna provides its own file storage implementation that may have `.db` suffix.

```python
storage = optuna.storages.JournalStorage(
    optuna.storages.JournalFileStorage("./optuna-journal.db"),
)
```

So Optuna Dashboard could not recognize the storage type from the file extension. Could you check the file header to determine which storages are given, as our browser-only version does?

https://github.com/optuna/optuna-dashboard/blob/9840f2ec03a532f17463b6e640dbd19bcf01d53a/standalone_app/src/components/StorageLoader.tsx#L36-L40

c-bata (changes requested):

optuna_dashboard/_storage_url.py
turbotimon: Thanks for pointing that out! In the docs i could only find examples with `.log` for the JournalStorage

turbotimon: I have changed it as you suggested

c-bata:
Let me assign @keisuke-umezawa as a reviewer. @turbotimon Could you fix a flake8 error?

turbotimon:
> Let me assign @keisuke-umezawa as a reviewer.
> 
> @turbotimon Could you fix a flake8 error?

Done!

PR #817, opened 2024-03-02
Use boolean state to stop API requests being issued multiple times

keisuke-umezawa:

## Reference Issues/PRs

NA

Optuna Dashboard performs near-real-time synchronization using API polling, but the current interval is 10 seconds.
If the number of trials is very large, it often does not return within 10 seconds.


porink0424 (approved):
LGTM!

PR #845, opened 2024-03-21
Move `optuna_dashboard/ts` to `optuna_dashboard_client` directory

c-bata:
## Contributor License Agreement

This repository (``optuna-dashboard``) and [Goptuna](https://github.com/c-bata/goptuna) share common code.
This pull request may therefore be ported to Goptuna.
Make sure that you understand the consequences concerning licenses and check the box below if you accept the term before creating this pull request.

* [x] I agree this patch may be ported to [Goptuna](https://github.com/c-bata/goptuna) by other Goptuna contributors.

## Reference Issues/PRs

None

## What does this implement/fix? Explain your changes.

Move `optuna_dashboard/ts` to `optuna_dashboard_client` direcotry.

```
$ mkdir optuna_dashboard_client
$ mv optuna_dashboard/ts optuna_dashboard_client/src
$ mv typescript_tests/ optuna_dashboard_client/tests/
$ mv ... & vim ...
```

Screenshot 2024-03-21 18 08 19

PR #847, opened 2024-03-22
Move config files to under the `optuna_dashboard` directory.

c-bata:

## Reference Issues/PRs

An alternative approach for #845.

Move config files to under the `optuna_dashboard` directory.


porink0424 (commented):
Looks almost good, but we also have to exclude `typescript_tests/**` from `biome.js`?

c-bata:
Thank you for your review! Good catch. I fixed the `biome.json`. Could you take a look again?

biome.json
```diff
@@ -4,8 +4,8 @@
 		"include": [
 			"optuna_dashboard/ts/**/*.ts",
 			"optuna_dashboard/ts/**/*.tsx",
-			"typescript_tests/**/*.ts",
-			"typescript_tests/**/*.tsx",
+			"optuna_dashboard/ts/tests/**/*.ts",
+			"optuna_dashboard/ts/tests/**/*.tsx",
```
porink0424, line 8: Maybe we can just delete these (I think these are already included into two lines above these).

c-bata: Ah, you are right 🙏

c-bata, line 8:
```suggestion
```

PR #846, opened 2024-03-21
Use TanStack Table instead of self-implemented DataGrid

keisuke-umezawa:
## Contributor License Agreement

This repository (``optuna-dashboard``) and [Goptuna](https://github.com/c-bata/goptuna) share common code.
This pull request may therefore be ported to Goptuna.
Make sure that you understand the consequences concerning licenses and check the box below if you accept the term before creating this pull request.

* [x] I agree this patch may be ported to [Goptuna](https://github.com/c-bata/goptuna) by other Goptuna contributors.

## Reference Issues/PRs

NA

## What does this implement/fix? Explain your changes.

Use TanStack Table instead of self-implemented DataGrid

PR #882, opened 2024-05-26
Use tanstack table in tslib

keisuke-umezawa:

## Reference Issues/PRs

NA

Tanstack table was introduced in optuna-dashboard by https://github.com/optuna/optuna-dashboard/pull/846.
This PR will also introduce it in tslib.

### DIFF
--- a/tslib/react/src/components/DataGrid.tsx
+++ b/tslib/react/src/components/DataGrid.tsx
@@ -119,17 +119,23 @@ function FilterMenu<T>({
 function DataGrid<T>({
   data,
   columns,
+  initialRowsPerPage,
 }: {
   data: T[]
   columns: ColumnDef<T>[]
+  initialRowsPerPage?: number
 }): React.ReactElement {
   const [sorting, setSorting] = React.useState<SortingState>([])
   const [columnFilters, setColumnFilters] = React.useState<ColumnFiltersState>(
     []
   )
+  const rowsPerPageOptions = [10, 50, 100, { label: "All", value: data.length }]
   const [pagination, setPagination] = React.useState<PaginationState>({
     pageIndex: 0,
-    pageSize: 50,
+    pageSize:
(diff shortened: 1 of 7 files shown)


tslib/react/src/components/DataGrid.tsx
```diff
@@ -12,372 +18,364 @@ import {
...
-          const field = columns[f.columnIdx].field
-          return row[field] === f.value
-        })
+  const [sorting, setSorting] = React.useState<SortingState>([])
+  const [columnFilters, setColumnFilters] = React.useState<ColumnFiltersState>(
+    []
+  )
+  const [pagination, setPagination] = React.useState<PaginationState>({
+    pageIndex: 0,
+    pageSize: 50,
```
porink0424, line 132: You set `pageSize` here 50 as constant, but how about receiving `initialRowsPerPage` as a prop and using that value here, as current `DataGrid` does?

tslib/react/src/components/TrialTable.tsx
```diff
@@ -1,153 +1,103 @@
...
+) => {
+  const rowValue = row.getValue(columnId) as string
+  return !filterValue.includes(rowValue)
+}
 
 export const TrialTable: FC<{
   study: Optuna.Study
-  initialRowsPerPage?: number
-}> = ({ study, initialRowsPerPage }) => {
+}> = ({ study }) => {
```
porink0424, line 24: Following the comment above, maybe you need to restore `initialRowsPerPage` prop here as well.
It would be also good in terms of not having to make destructive changes to the existing interface.

porink0424 (changes requested):
Thank you for your PR! Looks almost good.
I left a few requests.

keisuke-umezawa:
@porink0424 
Thank you for reviewing. I restored initialRowsPerPage.

porink0424 (approved):
LGTM👍

PR #906, opened 2024-07-19
Minor fix date axis in Timeline Chart for consistency with other charts

pandegaabyan:

## Reference Issues/PRs

N/A

The datetime or x-axis in the Timeline Chart is not consistent with datetime values in other places (History Chart and Trial detail). Given a timestamp from the database, the datetime axis in the Timeline Chart uses the UTC value, while other datetime values use the local value. If this is expected, I guess there should be a note or description in the chart.

optuna - timeline chart inconsistent date axis

I fixed it by simply removing the `.toISOString()`. Thus, instead of receiving a datetime string in UTC, Plotly receives a Date object. According to the [Plotly docs](https://plotly.com/javascript/reference/layout/xaxis/#layout-xaxis-range), the x-axis range supports `Date` objects. I have also successfully tested my fix to ensure that it is working.

### DIFF
--- a/optuna_dashboard/ts/components/GraphTimeline.tsx
+++ b/optuna_dashboard/ts/components/GraphTimeline.tsx
@@ -157,7 +157,7 @@ const plotTimeline = (
     xaxis: {
       title: "Datetime",
       type: "date",
-      range: [minDatetime.toISOString(), maxDatetime.toISOString()],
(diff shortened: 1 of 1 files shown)


c-bata (approved):
Looks perfect to me 💯  Thank you for your contribution!

Issue #911, opened 2024-07-27
CSV ArtifactViewer whites out when clicking a square


nabenabe0928:
### Description

When I upload [this CSV file](https://github.com/user-attachments/files/16398586/demo-dataset.csv) as an artifact to a study and click the square (see Figure 1) at the artifact, it whites out.

![image](https://github.com/user-attachments/assets/2297d94c-cbb8-4005-9994-9460c3b839c6)
Figure 1. The icon of the artifact

![image](https://github.com/user-attachments/assets/f0727728-751c-4faf-8451-1e77d8242fcf)
Figure 2. The dashboard page

The page in Figure 2. whites out when clicking the square button.
The URL does not change, the `Go Back` button does not resolve the problem. Refreshing the page fixes the problem.

### How to Reproduce

1. Start dashboard
2. Upload [this CSV file](https://github.com/user-attachments/files/16398586/demo-dataset.csv) to the dashboard
4. Click the button
5. White out

### Python version

3.9

### Optuna version

4.0.0b0

### optuna-dashboard version or git revision

The  HEAD of the main branch 

### Web browser

Google Chrome

c-bata:
@keisuke-umezawa Could you take a look at this issue?

c-bata:
I also notice an another issue that `TableArtifactViewer` is not enabled when using trial artifacts.
https://github.com/optuna/optuna-dashboard/pull/877

keisuke-umezawa:
@c-bata ok! I will handle it.

nabenabe0928:
@keisuke-umezawa @c-bata 

Replacing an empty string with a random string resolved the issue:

```tsx
    return keys.map((key) => ({
      header: key || " ",
      accessorFn: (info: Data) =>
        typeof info[key] === "object" ? JSON.stringify(info[key]) : info[key],
      enableSorting: true,
      enableColumnFilter: false,
    }))
```

https://github.com/optuna/optuna-dashboard/blob/main/optuna_dashboard/ts/components/Artifact/TableArtifactViewer.tsx#L58

c-bata:
Fixed in #913.

PR #913, opened 2024-07-31
Fix an error in TableArtifactViewer when keys have an empty string

nabenabe0928:

## Reference Issues/PRs

This issue relates to
- https://github.com/optuna/optuna-dashboard/issues/911

In my implementation, I replaced an empty string in keys with `" "`, which circumvents the error.

### DIFF
--- a/optuna_dashboard/ts/components/Artifact/TableArtifactViewer.tsx
+++ b/optuna_dashboard/ts/components/Artifact/TableArtifactViewer.tsx
@@ -55,7 +55,7 @@ export const TableArtifactViewer: React.FC<TableArtifactViewerProps> = (
     })
     const keys = Array.from(unionSet)
     return keys.map((key) => ({
-      header: key,
+      header: key || " ",
       accessorFn: (info: Data) =>
         typeof info[key] === "object" ? JSON.stringify(info[key]) : info[key],
       enableSorting: true,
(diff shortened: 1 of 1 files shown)


c-bata (commented):
@nabenabe0928 Thank you for your pull request! The changes look good, but the cause of issue #911 appears to be that `Papa.parse()` does not skip empty lines by default, leading to the result shown below:

Screenshot 2024-07-31 16 43 54

Could you pass the `skipEmptyLines: true` option to `Papa.parse()`?

```diff
$ git diff
diff --git a/optuna_dashboard/ts/components/Artifact/TableArtifactViewer.tsx b/optuna_dashboard/ts/components/Artifact/TableArtifactViewer.tsx
index a9a2414..644dbff 100644
--- a/optuna_dashboard/ts/components/Artifact/TableArtifactViewer.tsx
+++ b/optuna_dashboard/ts/components/Artifact/TableArtifactViewer.tsx
@@ -143,6 +143,7 @@ const loadCSV = (props: TableArtifactViewerProps): Promise<Data[]> => {
     Papa.parse(props.src, {
       header: true,
       download: true,
+      skipEmptyLines: true,
       complete: (results: Papa.ParseResult<Data>) => {
         resolve(results?.data)
       },
```

https://www.papaparse.com/docs

nabenabe0928:
@c-bata I tried your modification, but did not work (still whites out...)

c-bata:
@nabenabe0928 Hmm... Could you please try the hard refresh? The patch seems to work fine on my environment actually.

Screenshot 2024-08-01 12 50 32

nabenabe0928:
image

image

I tried ctrl+f5 and the source is updated as highlighted.
However, I still get the same error, but I do not know why.

c-bata (approved):
Thank you for your efforts to reproduce! It appears I misunderstood the cause of issue #911. Without the changes made in this PR, I also encountered the error you ran into.

Regarding the patch I suggested, it should be applied. However, it can be addressed in a follow-up. LGTM.

PR #917, opened 2024-08-03
Sort artifact cards by filename

nabenabe0928:

## Reference Issues/PRs

image

In the current implementation, the artifact cards are sorted by `artifact_id`, which is only used internally.
It is more intuitive to sort the cards either by the time added or filename.
As artifact meta does  not include the time when it was added, I simply sorted by filename.

With this change, artifact cards will be sorted by filename.

image


porink0424 (approved):
LGTM👍

PR #972, opened 2024-09-26
Add constraint null checking in graph components

hrntsm:

In the same study, when a trial with constraints and a trial without constraints were mixed together, an exception was generated when drawing graphs.

Specifically, the null check was added because a Trial without constraints would be null.
In the case of null, Trial is feasible.

### DIFF
--- a/optuna_dashboard/ts/components/GraphContour.tsx
+++ b/optuna_dashboard/ts/components/GraphContour.tsx
@@ -287,7 +287,10 @@ const plotContour = (
 
   filteredTrials.forEach((trial, i) => {
     if (xAxis.values[i] && yAxis.values[i] && trial.values) {
-      if (trial.constraints.every((c) => c <= 0)) {
+      if (
+        trial.constraints === null ||
+        trial.constraints.every((c) => c <= 0)
+      ) {
         feasibleXY.add(xValues.length)
       }
       const xValue = xAxis.values[i] as string | number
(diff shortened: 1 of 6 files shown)


porink0424 (changes requested):
Thank you for the PR!

Currently, type `Trial.constraints` is implemented in a way that `null` cannot be included (also, in the `serialize_frozen_trial` function of `optuna_dashboard/_serializer.py`, `constraints` are created like `"constraints": trial_system_attrs.get(CONSTRAINTS_KEY, [])`, which means I don't think `null` will be included). So I'm not sure if it's really necessary to consider the possibility of `null` here.

If it is indeed necessary, could you provide an example code that produces such a situation?

hrntsm:
@porink0424 

Thank you for checking it out.
I was not able to reproduce the problem, so I reviewed the log file and found that it was caused by the following null included in the file.

```
{"op_code": 9, "worker_id": "5609884b-ac1d-4851-854b-c2baa6e293e4-5712", "trial_id": 5, "system_attr": {"constraints": null}}
```

However, looking at the following Optuna implementation, it seems that it would result in an error in case of null.
Therefore, this kind of result should not occur.
As I can't understand the cause, I think I will close this PR for now.

https://github.com/optuna/optuna/blob/8476b460fb4ad79e9051a3b7ac49629d0e389ac9/optuna/samplers/_base.py#L244

PR #1004, opened 2024-12-25
Fix the problem that SAVE button and CLOSE button are invisible when editing Note

fusawa-yugo:

* [X] I agree this patch may be ported to [Goptuna](https://github.com/c-bata/goptuna) by other Goptuna contributors.

## Reference Issues/PRs

The problem is that we can't see SAVE button and CLOSE button when we edit markdown of Note. 

The cause is textarea of markdown field expands infinitely (but I'm not sure why this happens).
So I set the height of the field using `calc`.

スクリーンショット 2024-12-25 16 31 55

### DIFF
--- a/optuna_dashboard/ts/components/Note.tsx
+++ b/optuna_dashboard/ts/components/Note.tsx
@@ -331,9 +331,9 @@ const MarkdownEditorModal: FC<{
             resize: "none",
             width: "100%",
             height: "100%",
-            "& .MuiInputBase-root": { 
-              height: "100%", 
-              alignItems: "start" 
+            "& .MuiInputBase-root": {
+              height: "100%",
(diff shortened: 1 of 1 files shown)


porink0424 (commented):
Thank you for the PR!
Locking the height is a reasonable temporary solution, but, based on my investigation, I found that removing the [`inputProps` from the TextField](https://github.com/optuna/optuna-dashboard/blob/main/optuna_dashboard/ts/components/Note.tsx#L336-L338) resolves the issue (It seems the duplicate height specification might be causing the problem).

Since `inputProps` is already deprecated, could you consider whether it's safe to exclude it altogether?

porink0424 (approved):
LGTM👍

PR #1022, opened 2025-01-29
Fix the structure of cache to avoid unexpected errors

fusawa-yugo:

The previous implementation used a cache with a module-global scope.
In this case, cache contamination occurs when you run multiple Optuna Dashboard servers with different storages within a single process. 

To address this, I made the cache an attribute of the Bottle object.
```
app = Bottle()
app._inmemory_cache = InMemoryCache()
```

### DIFF
--- a/optuna_dashboard/_importance.py
+++ b/optuna_dashboard/_importance.py
@@ -98,11 +98,11 @@ def _get_param_importances(
 
 
 def get_param_importance_from_trials_cache(
+    inmemory_cache: InMemoryCache,
     storage: BaseStorage,
     study_id: int,
     objective_id: int,
     trials: list[FrozenTrial],
-    inmemory_cache: InMemoryCache,
 ) -> list[ImportanceType]:
     completed_trials = [t for t in trials if t.state == TrialState.COMPLETE]
     n_completed_trials = len(completed_trials)
--- a/optuna_dashboard/_importance.py
+++ b/optuna_dashboard/_importance.py
@@ -11,7 +11,8 @@ from optuna.storages import BaseStorage
 from optuna.study import Study
 from optuna.trial import FrozenTrial
 from optuna.trial import TrialState
-from optuna_dashboard._cached_extra_study_property import get_cached_extra_study_property
+from optuna_dashboard._inmemory_cache import get_cached_extra_study_property
+from optuna_dashboard._inmemory_cache import InMemoryCache
 
 
 _logger = logging.getLogger(__name__)
@@ -97,7 +98,11 @@ def _get_param_importances(
 
 
 def get_param_importance_from_trials_cache(
-    storage: BaseStorage, study_id: int, objective_id: int, trials: list[FrozenTrial]
+    storage: BaseStorage,
+    study_id: int,
+    objective_id: int,
+    trials: list[FrozenTrial],
+    inmemory_cache: InMemoryCache,
 ) -> list[ImportanceType]:
     completed_trials = [t for t in trials if t.state == TrialState.COMPLETE]
     n_completed_trials = len(completed_trials)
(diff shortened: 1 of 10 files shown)


optuna_dashboard/_importance.py
```diff
@@ -97,7 +98,11 @@ def _get_param_importances(
 
 
 def get_param_importance_from_trials_cache(
-    storage: BaseStorage, study_id: int, objective_id: int, trials: list[FrozenTrial]
+    storage: BaseStorage,
+    study_id: int,
+    objective_id: int,
+    trials: list[FrozenTrial],
+    inmemory_cache: InMemoryCache,
```
porink0424, line 105: It would be better to place the argument `inmemory_cache` at the beginning to match the other functions.

optuna_dashboard/_inmemory_cache.py
```diff
@@ -41,6 +41,48 @@ def get_cached_extra_study_property(
...
+    def __init__(self) -> None:
+        self._cached_extra_study_property_cache: dict[int, "_CachedExtraStudyProperty"] = {}
+        self._cached_extra_study_property_cache_lock = threading.Lock()
+        self._trials_cache: dict[int, list[FrozenTrial]] = {}
+        self._trials_cache_lock = threading.Lock()
+        self._trials_last_fetched_at: dict[int, datetime] = {}
+
+    def clear(self) -> None:
+        self._cached_extra_study_property_cache.clear()
+        self._trials_cache.clear()
```
porink0424, line 83: Should we also clear `self._trials_last_fetched_at` here?

optuna_dashboard/_storage.py
```diff
@@ -17,33 +16,6 @@
 trials_last_fetched_at: dict[int, datetime] = {}
```
porink0424, line 16: Are these 3 lines no longer needed?

optuna_dashboard/_inmemory_cache.py
```diff
@@ -41,6 +41,48 @@ def get_cached_extra_study_property(
         )
 
 
+def get_trials(
```
porink0424, line 44: IMO, `get_trials` primarily operates on storage, with the cache serving a secondary role. So keeping this function in `_storage.py` as it was originally seems clearer to me. What do you think?

porink0424 (changes requested):
Thank you for the PR, I've left some comments. PTAL.

optuna_dashboard/_inmemory_cache.py
fusawa-yugo: I agree with you.
Thank you!

fusawa-yugo:
@porink0424 
Thank you for your feedbacks!!
Could you review these again?

optuna_dashboard/_inmemory_cache.py
```diff
@@ -41,6 +39,20 @@ def get_cached_extra_study_property(
...
+        self._cached_extra_study_property_cache: dict[int, "_CachedExtraStudyProperty"] = {}
+        self._cached_extra_study_property_cache_lock = threading.Lock()
+        self._trials_cache: dict[int, list[FrozenTrial]] = {}
+        self._trials_cache_lock = threading.Lock()
+        self._trials_last_fetched_at: dict[int, datetime] = {}
+
+    def clear(self) -> None:
+        self._cached_extra_study_property_cache.clear()
+        self._trials_cache.clear()
+        self._trials_last_fetched_at.clear()
```
porink0424, line 53: Do we need the lock here as well?
```suggestion
        with self._cached_extra_study_property_cache_lock:
            self._cached_extra_study_property_cache.clear()
        with self._trials_cache_lock:
            self._trials_cache.clear()
            self._trials_last_fetched_at.clear()
```

porink0424 (commented):
Thank you, it's almost LGTM—I will merge this PR once one comment I added is addressed.

fusawa-yugo:
@porink0424 
Thank you for your reviews!!
I fixed it.

porink0424 (approved):
LGTM!

PR #1076, opened 2025-05-27
CLI enable sqlite path

turbotimon:

## Reference Issues/PRs

Follow-up: #799

Since #800  the cli provides a clear error message when trying to start optuna-dashboard with a file path to a sqlite database. But actually, since optuna-dashboard knows it is a sqlite database, it would be more user-friendly to just open it. This has the additional benefit, that one can use the path auto-completion in bash. Specifying the url still works.

In short:

- Enable passing a sqlite db as path, e.g. `optuna-dashboard ./db.sqplite3`
- Improving help texts for the cli
- Adding to the quick start

### DIFF
--- a/docs/getting-started.rst
+++ b/docs/getting-started.rst
@@ -92,6 +92,16 @@ you can use the command line interface like ``optuna-dashboard <STORAGE_URL>``.
    Listening on http://localhost:8080/
    Hit Ctrl-C to quit.
 
+
+Alternatively, for local SQLite3 databases, you can provide the file path directly:
+
+.. code-block:: console
+
+   $ optuna-dashboard ./db.sqlite3
+   Listening on http://localhost:8080/
+   Hit Ctrl-C to quit.
+
+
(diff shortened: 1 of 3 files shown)


c-bata:
@turbotimon Thank you for your effort! We've actually considered this in the past, but we would like to keep the `--storage` argument in optuna-dashboard as consistent as possible with the `--storage` argument in the Optuna CLI. When a user specifies something like `optuna create-study --storage journal.db`, Optuna cannot determine whether it should create an SQLite3 database file or a JournalStorage (with JournalFileBackend) file.

We completely understand the motivation behind this proposal. However, for the reasons above, we will be closing this PR. Thanks again 🥇.

turbotimon:
@c-bata I kind of understand, but I want to point out that it isn't a problem in my opinion:

- It would be only for `optuna-dashboard` not for other tools
- It is consisted as you still can use the url path and a `optuna-dashboard journal.db` (from your example) would also still work
- Have you considered at least taking my suggestion to improves the CLI help?

Issue #1105, opened 2025-07-15
Design Document: LLM Integration


c-bata:
## Overview

* Introducing LLM integration functionality to the Optuna Dashboard.
* Our first implementation will allow users to filter trials displayed in the Optuna Trial list or on Plotly charts using natural language. While this may seem like a relatively minor update—just adding a single text input field to the UI—it actually provides a powerful feature that existing Optuna Dashboard users will find intuitive to use—and surprisingly effective once experienced.

<img width="988" height="664" alt="Image" src="https://github.com/user-attachments/assets/957956fb-734d-465e-9295-fd4ab57f9f1b" />

### Internal Processing Flow

```mermaid
sequenceDiagram
    participant User
    participant WebBrowser as Web Browser
    participant APIServer as API Server
    participant LLMProvider as LLM Provider

    User->>WebBrowser: Enter query in form
    WebBrowser->>APIServer: HTTP GET Request /api/llm/trial_filter_query
    APIServer->>LLMProvider: Prompt
    LLMProvider->>APIServer: JavaScript function (string)
    APIServer->>WebBrowser: JavaScript function (string)
    WebBrowser->>User: Show execution confirmation dialog
    User->>WebBrowser: "Allow" or "Reject"
    WebBrowser->>WebBrowser: Execute eval()
    note right of WebBrowser: eval() is executed in a sandboxed iframe environment
    WebBrowser->>User: Filtered trial list
```

Depending on the LLM's capabilities, the system may occasionally generate JavaScript functions that cannot be executed. To address this, the diagram omits details but includes error messaging to the API server while repeatedly regenerating the JavaScript function and eval() execution up to approximately five times. When using GPT 4o-mini, success is typically achieved within 2-3 attempts in most cases.

### Executing eval() in a Sandboxed iframe Environment
To mitigate potential side effects of eval(), such as access to the host environment or document context, we execute it inside a sandboxed iframe. Trial list data and JavaScript functions (as strings) are passed to the iframe via postMessage(), and eval() is executed in the isolated context to return the filtered trial list.

### User Confirmation Dialog
As a simple preventive measure, we display a confirmation dialog before executing any JavaScript functions.

<img width="798" height="450" alt="Image" src="https://github.com/user-attachments/assets/4ea3d2e9-a749-4c77-ba13-235dcb88f6a7" />

* The JavaScript function can be displayed by clicking the "Expand" button, similar to how it functions in Claude Desktop.
[...]

PR #1132, opened 2025-08-06
Add TOML config file.

toshihikoyanase:

## Reference Issues/PRs

Related to https://github.com/optuna/optuna-dashboard/issues/1105

- Add support for `--from-config` command-line option to specify dashboard configuration in TOML format
- Users can specify various configurations such as `storage` and `artifact_store`, as well as LLM configuration to activate LLM integration (#1105)

### DIFF
--- a/optuna_dashboard/_cli.py
+++ b/optuna_dashboard/_cli.py
@@ -139,7 +139,9 @@ def main() -> None:
         config = {}
 
     if args.storage is None:
-        raise ValueError("Storage URL must be specified as the first argument or via --from-config.")
+        raise ValueError(
+            "Storage URL must be specified as the first argument or via --from-config."
+        )
 
     storage: BaseStorage
     storage = get_storage(args.storage, storage_class=args.storage_class)
--- a/optuna_dashboard/_cli.py
+++ b/optuna_dashboard/_cli.py
@@ -139,7 +139,7 @@ def main() -> None:
         config = {}
 
     if args.storage is None:
-        raise ValueError("Storage URL must be specified. Use --storage or --from-config.")
+        raise ValueError("Storage URL must be specified as the first argument or via --from-config.")
 
     storage: BaseStorage
     storage = get_storage(args.storage, storage_class=args.storage_class)
--- a/optuna_dashboard/_cli.py
+++ b/optuna_dashboard/_cli.py
@@ -4,6 +4,7 @@ import argparse
 import os
 from socketserver import ThreadingMixIn
 import sys
+from typing import Any
 from typing import TYPE_CHECKING
 from wsgiref.simple_server import make_server
 from wsgiref.simple_server import WSGIServer
@@ -15,6 +16,7 @@ from optuna.storages import RDBStorage
 
 from . import __version__
 from ._app import create_app
+from ._config import load_config_from_toml
 from ._sql_profiler import register_profiler_view
 from ._storage_url import get_storage
 
@@ -23,6 +25,8 @@ if TYPE_CHECKING:
     from typing import Literal
 
     from optuna.artifacts._protocol import ArtifactStore
+
+    from .llm.provider import LLMProvider
 
 
 DEBUG = os.environ.get("OPTUNA_DASHBOARD_DEBUG") == "1"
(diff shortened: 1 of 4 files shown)


optuna_dashboard/_config.py
```diff
@@ -0,0 +1,33 @@
...
+import sys
+from typing import Any
+
+if sys.version_info >= (3, 11):
+    import tomllib
+else:
+    try:
+        import tomli as tomllib
+    except ImportError:
+        raise ImportError("tomli is required for Python < 3.11. Install with: pip install tomli")
```
c-bata, line 12: Can we simply import tomli since it is specified as the mandatory dependency in pyproject.toml?

```suggestion
    import tomli as tomllib
```

python_tests/test_config.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def test_empty_file() -> None:
+    config_content = ""
+    with tempfile.NamedTemporaryFile(mode="w", suffix=".toml") as f:
+        f.write(config_content)
+        f.flush()
+
+        config = load_config_from_toml(f.name)
+        assert config == {}
```
c-bata, line 89: These test scenarios look like mostly testing the behavior for toml library itself, rather than testing our `load_config_from_toml` function. What do you think?

optuna_dashboard/_cli.py
```diff
@@ -83,9 +87,72 @@ def auto_select_server(
...
+
+        artifact_store = GCSArtifactStore(**config["artifact_store"]["gcs"])
+    elif "filesystem" in config["artifact_store"]:
+        from optuna.artifacts import FileSystemArtifactStore
+
+        artifact_store = FileSystemArtifactStore(**config["artifact_store"]["filesystem"])
+    else:
+        raise ValueError("Unsupported artifact store configuration.")
+
+    return artifact_store
```
c-bata, line 144: It might be good to move these functions to `_config.py` to avoid becoming `_cli.py` large. Then, we can write unit tests like below:

```python
def test_create_artifact_store_from_config() -> None:
    config_content = """
[optuna_dashboard]
storage = "sqlite:///test.db"

[artifact_store.filesystem]
base_path = "/tmp/artifacts"
"""
    with tempfile.NamedTemporaryFile(mode="w", suffix=".toml") as f:
        f.write(config_content)
        f.flush()

        config = load_config_from_toml(f.name)
    artifact_store = create_artifact_store_from_config(config)
    assert isinstance(artifact_store, FileSystemArtifactStore)
```

c-bata (commented):
Thank you for your pull request. I left some suggestions.

python_tests/test_config.py
toshihikoyanase: I agree.
On the other hand, since `load_config_from_toml` is loading configurations while switching between two different toml libraries, we may still conduct load testing for some toml files.
How about eliminating the test scenario and reducing the number of test cases?

toshihikoyanase: Ah, [this kind of tests](https://github.com/optuna/optuna-dashboard/pull/1132#discussion_r2259735187) can remove these test cases.
Although it might be a bit of integration tests rather than unit tests.

optuna_dashboard/_cli.py
toshihikoyanase: I moved the `create_artifact_store_from_config` and `create_llm_provider_from_config` to `_config.py` and added unit tests for them in the commit https://github.com/optuna/optuna-dashboard/pull/1132/commits/24ad2ba193dcf0bd392c7873326478a2363a2aa9.

optuna_dashboard/_cli.py
```diff
@@ -107,21 +116,49 @@ def main() -> None:
         help="directory to store artifact files",
         default=None,
     )
+    parser.add_argument(
+        "--from-config",
+        help="configuration file in TOML format",
+        type=str,
+        default=None,
+    )
```
c-bata, line 124: I was initially considering automatically reading `optuna-dashboard.toml` if it exists, but I’m wondering if this approach might be confusing for the user. What do you think about this idea?

c-bata (commented):
Changes look almost good to me. I left one queestion though.

optuna_dashboard/_cli.py
toshihikoyanase: Searching for `optuna-dashboard.toml` in the current directory may cause unexpected behavior because `cd`  can affect the search path.

`grafana` follows [this configuration search priority](https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#configuration-file-location) and loads configuration files from conf/ in the current directory (custom.ini, default.ini):

1. Explicitly specified via the --config option
2. <Current Working Directory>/conf/custom.ini, <Current Working Directory>/conf/default.ini

Prometheus does not load default configuration files by default and requires explicit specification using the `--config.file` option.

From this analysis, both scenarios are possible: either loading the default files in the current directory or omitting default values.

IMHO, we're currently implementing an LLM, but designing it to prevent implicit activation would be safe—it would avoid accidental billing issues.

optuna_dashboard/_cli.py
```diff
@@ -107,21 +116,49 @@ def main() -> None:
...
+        args.storage_class = cli_config.get("storage_class", args.storage_class)
+        args.port = cli_config.get("port", args.port)
+        args.host = cli_config.get("host", args.host)
+        args.server = cli_config.get("server", args.server)
+        args.artifact_dir = cli_config.get("artifact_dir", args.artifact_dir)
+    else:
+        config = {}
+
+    if args.storage is None:
+        raise ValueError("Storage URL must be specified. Use --storage or --from-config.")
```
c-bata, line 142: I've noticed that `--storage` does not exist.

```
$ uv run optuna-dashboard --storage sqlite:///db.sqlite3
Traceback (most recent call last):
  File "/Users/c-bata/go/src/github.com/optuna/optuna-dashboard/.venv/bin/optuna-dashboard", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/c-bata/go/src/github.com/optuna/optuna-dashboard/optuna_dashboard/_cli.py", line 142, in main
    raise ValueError("Storage URL must be specified. Use --storage or --from-config.")
ValueError: Storage URL must be specified. Use --storage or --from-config.
```

Could you update the error message like below?

```suggestion
        raise ValueError("Storage URL must be specified as the first argument or via --from-config.")
```

optuna_dashboard/_cli.py
c-bata: Thank you for your opinion. It sounds make sense.

c-bata (commented):
Thank you for the update. I left one additional request to change 🙏

optuna_dashboard/_cli.py
toshihikoyanase: You're right. Thank you!

c-bata (approved):
LGTM.

PR #1136, opened 2025-08-21
Add documentation for TOML configuration files

toshihikoyanase:

## Reference Issues/PRs

A follow-up PR for #1132.

* Add a new page to explain TOML configuration files.
* Add description of `--from-config` option to CLI section of the getting started document.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -27,7 +27,6 @@ Configuration file settings are overridden by command-line arguments.
     port = 8080
     host = "127.0.0.1"
     server = "auto"
-    artifact_dir = "/path/to/artifacts"
 
 LLM Integration
 ~~~~~~~~~~~~~~~
@@ -40,7 +39,7 @@ OpenAI
 Configure OpenAI or OpenAI-compatible API.
 
 The ``llm.openai`` configuration specifies the model name and API type.
-When ``use_chat_completions_api`` is set to ``true``, the `OpenAI chat completions API <https://platform.openai.com/docs/api-reference/chat>`__ will be used; setting it to ``false`` will use the `Responses API <https://platform.openai.com/docs/api-reference/responses/create>`__.
+When ``use_chat_completions_api`` is set to ``true``, the `OpenAI Chat Completions API <https://platform.openai.com/docs/api-reference/chat>`__ will be used; setting it to ``false`` will use the `Responses API <https://platform.openai.com/docs/api-reference/responses/create>`__.
 
 The ``llm.openai.client`` configuration specifies the OpenAI API key and endpoint, which are passed to the constructor of the `openai.OpenAI <https://github.com/openai/openai-python>`__ class.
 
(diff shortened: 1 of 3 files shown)


docs/configuration.rst
```diff
@@ -0,0 +1,137 @@
...
+
+.. code-block:: toml
+
+    [optuna_dashboard]
+    storage = "sqlite:///example.db"
+    storage_class = "RDBStorage"
+    port = 8080
+    host = "127.0.0.1"
+    server = "auto"
+    artifact_dir = "/path/to/artifacts"
```
c-bata, line 30: I've noticed that there are two ways to enable FileSystemArtifact.

```
[optuna_dashboard]
artifact_dir = "/path/to/artifacts"

[artifact_store.filesystem]
base_path = "/path/to/artifacts"
```

In my opinion, we should avoid providing the `optuna_dashboard.artifact_dir` option, since the latter approach is more consistent and flexible. How about removing this line from the documentation? Later, we can also update the implementation in a future pull request:

https://github.com/optuna/optuna-dashboard/blob/1ecfa1b1c37a7c91da41206dd4c3aec8eaa1aab3/optuna_dashboard/_cli.py#L137

```suggestion
```

docs/configuration.rst
```diff
@@ -0,0 +1,137 @@
...
+
+To enable LLM Integration, configure the LLM Provider.
+
+OpenAI
+^^^^^^
+
+Configure OpenAI or OpenAI-compatible API.
+
+The ``llm.openai`` configuration specifies the model name and API type.
+When ``use_chat_completions_api`` is set to ``true``, the `OpenAI chat completions API <https://platform.openai.com/docs/api-reference/chat>`__ will be used; setting it to ``false`` will use the `Responses API <https://platform.openai.com/docs/api-reference/responses/create>`__.
```
c-bata, line 43: This is very nit-picky, but OpenAI's official documentation refers to it as "Chat Completions API" (both words capitalized).

> The Chat Completions API endpoint will generate a model response from a list of messages comprising a conversation.
> https://platform.openai.com/docs/api-reference/chat

```suggestion
When ``use_chat_completions_api`` is set to ``true``, the `OpenAI Chat Completions API <https://platform.openai.com/docs/api-reference/chat>`__ will be used; setting it to ``false`` will use the `Responses API <https://platform.openai.com/docs/api-reference/responses/create>`__.
```

c-bata (commented):
Looks almost good to me. I left two minor comments.

docs/configuration.rst
toshihikoyanase: Thank you for pointing it out! Sounds reasonable.

docs/configuration.rst
toshihikoyanase: Woops! Thank you.

c-bata (approved):
LGTM!

PR #1142, opened 2025-08-26
Add LLM filter to history

y0z:

This PR introduces an LLM filter function to the history section.

image

### DIFF
--- a/optuna_dashboard/ts/components/GraphHistory.tsx
+++ b/optuna_dashboard/ts/components/GraphHistory.tsx
@@ -1,9 +1,24 @@
+import ClearIcon from "@mui/icons-material/Clear"
+import FilterListIcon from "@mui/icons-material/FilterList"
 import { useTheme } from "@mui/material"
+import {
+  Box,
+  Button,
+  Card,
+  CardContent,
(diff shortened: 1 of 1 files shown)


optuna_dashboard/ts/components/GraphHistory.tsx
```diff
@@ -21,14 +71,59 @@ export const GraphHistory: FC<{
...
-    <PlotHistory
-      studies={studies}
-      logScale={logScale}
-      includePruned={includePruned}
-      colorTheme={colorTheme}
-      linkURL={linkURL}
-      router={navigate}
-      selectedTrials={selectedTrials}
-    />
+    <Box
```
toshihikoyanase, line 74: The current implementation wraps the `PlotHistory` object in a box and places the filter text field within it.
As a result, the history plot on the selection page also includes the text field.
The selection page has both a graph and a trial list, so users might expect synchronization between them. What do you think of it?

image

toshihikoyanase (commented):
Thank you for your PR. I have a comment about the design choice (component-wise implementation vs page-wise implementation). So, please take a look.

PR #1144, opened 2025-08-28
Code fix and add tests for config priority

y0z:

Based on [the documentation](https://optuna-dashboard.readthedocs.io/en/latest/configuration.html#priority-order), the config priority is CLI options > config file > default values. I updated the implementation to align with this priority order and added tests to verify the priority.

image


toshihikoyanase:
Thanks for fixing the priority issues!
I'll check the changes.

optuna_dashboard/_cli.py
```diff
@@ -129,11 +133,17 @@ def main() -> None:
...
+        args.storage_class = (
+            args.storage_class
+            if args.storage_class is not None
+            else cli_config.get("storage_class")
+        )
+        args.port = args.port if args.port is not None else cli_config.get("port", DEFAULT_PORT)
+        args.host = args.host if args.host is not None else cli_config.get("host", DEFAULT_HOST)
+        args.server = (
+            args.server if args.server is not None else cli_config.get("server", DEFAULT_SERVER)
+        )
```
toshihikoyanase, line 146: Thank you for fixing the priority issue!

Not a strong opinion, but the `A or B` pattern might simplify the code, and developers might be able to follow the logic easily. We should be careful when the case of `A = 0` or `A = '' ', but we can ignore such cases in this situation. Also, we can see this pattern in existing code, such as https://github.com/optuna/optuna-dashboard/blob/3995d19f90bfcd3356d7721d4b4055144ffb33d0/optuna_dashboard/_app.py#L204 and https://github.com/optuna/optuna-dashboard/blob/3995d19f90bfcd3356d7721d4b4055144ffb33d0/optuna_dashboard/llm/_prompt_templates.py#L112, so I don't think it violates the coding convention of this repository. What do you think?

```suggestion
        args.storage = args.storage or cli_config.get("storage")
        args.storage_class = args.storage_class or cli_config.get("storage_class")
        args.port = args.port or cli_config.get("port", DEFAULT_PORT)
        args.host = args.host or cli_config.get("host", DEFAULT_HOST)
        args.server = args.server or cli_config.get("server", DEFAULT_SERVER)
```

toshihikoyanase (commented):
Thank you for your PR. The logic seems great to me.
I'll share my comments on coding style and code location.

python_tests/test_config.py
```diff
@@ -82,3 +84,118 @@ def test_create_llm_provider_from_config() -> None:
         config = load_config_from_toml(f.name)
     llm_provider = create_llm_provider_from_config(config)
     assert isinstance(llm_provider, OpenAI)
+
+
+def test_cli_args_override_config_file() -> None:
```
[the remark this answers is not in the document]
y0z: Thank you. I applied your suggestion.

python_tests/test_cli.py
```diff
@@ -0,0 +1,120 @@
...
+            "optuna_dashboard._cli.create_app"
+        ) as mock_create_app, patch(
+            "optuna_dashboard._cli.run_debug_server"
+        ) as mock_run_debug_server, patch("optuna_dashboard._cli.DEBUG", True):
+            # Setup mocks to prevent actual execution.
+            mock_storage = MagicMock()
+            mock_get_storage.return_value = mock_storage
+            mock_app = MagicMock()
+            mock_create_app.return_value = mock_app
+            mock_run_debug_server.return_value = None  # Prevent server startup.
```
toshihikoyanase, line 31: Executing `main()` in `cli.py` involves patching many components, making the test code complicated. From a maintainability perspective, this could potentially increase future engineering costs.

How about adding a helper function or data class that determines values according to the priority of CLI options and configuration files?

toshihikoyanase (commented):
Thanks for the update. Starting from your new test code, I tried different test approaches, but I couldn't make the code simple.

Embedding priority determination within the CLI may require writing integration tests. How about extracting the priority determination logic into a function and testing that component separately?

PR #1208, opened 2025-12-26
Fix: Correctly sort numeric values in TrialTable (params and user_attrs)

### LINKED ISSUE #1207: Incorrect sorting of negative numeric values in Parameters and User Attributes
### Description

When Parameters (`params`) or User Attributes (`user_attrs`) contain numeric values, they are serialized as strings by the backend. The frontend `TrialTable` currently relies on a default sorting mechanism (likely "natural" or alphanumeric sorting).

While this mechanism correctly handles positive integers (e.g., sorting `2` before `10`), it fails for **negative numbers**. The sorting logic appears to treat the minus sign as a character and sorts based on the magnitude of the number following it.

**Current Behavior:**

* Positive numbers sort correctly: `2, 10, 100`
* Negative numbers sort incorrectly (inverted order): `-1, -2, -10` (instead of `-10, -2, -1`).

**Expected Behavior:**
The dashboard should detect if the column (Parameter or User Attribute) contains numeric values and apply a true numeric sort function that correctly accounts for negative signs.

### How to Reproduce

1. Create a Python script (`reproduce_issue.py`) to generate a study with negative values in both parameters and user attributes:
```python
import optuna

def objective(trial):
    # 1. Create a parameter with a negative range
    x = trial.suggest_int("x_param", -10, 10)

    # 2. Set the same value as a user attribute
    trial.set_user_attr("x_attr", x)

    return x ** 2

if __name__ == "__main__":
    study = optuna.create_study(storage="sqlite:///db.sqlite3", study_name="sort_bug_test")
    study.optimize(objective, n_trials=20)
```

2. Run the script: `python reproduce_issue.py`
3. Launch the dashboard: `optuna-dashboard sqlite:///db.sqlite3`
4. Open the **Trials** tab (Trial Table).
5. Click the header for **Param x_param** and **UserAttribute x_attr** to sort them.
6. Observe that for both columns, negative numbers are sorted by magnitude.

### Python version

Python 3.12.12

### Optuna version

4.4.0

0.20.0

### Web browser

Microsoft Edge Beta

MiracleInvoker:

## Reference Issues/PRs

Fixes #1207  

Previously, the default sorting treated these values as strings/text. This caused negative numbers to appear in reverse order (e.g., `-1` appearing before `-2` in ascending sort), as they were sorted by character rather than numeric value.

I have added a `sortingFn` to the column definitions for both Params and User Attributes in `TrialTable.tsx`. This function safely converts values to numbers for comparison, ensuring that negative numbers, scientific notation, and mixed types sort in the expected numeric order.

### DIFF
--- a/tslib/react/src/components/TrialTable.tsx
+++ b/tslib/react/src/components/TrialTable.tsx
@@ -127,6 +127,24 @@ export const TrialTable: FC<{
             header: `Param ${s.name}`,
             enableSorting: sortable,
             sortUndefined: "last",
+            sortingFn: (rowA, rowB, columnId) => {
+              const valA = rowA.getValue(columnId) as string | null
+              const valB = rowB.getValue(columnId) as string | null
+              const numA = Number(valA)
+              const numB = Number(valB)
+
+              if (
+                valA !== null &&
+                valB !== null &&
+                !Number.isNaN(numA) &&
+                !Number.isNaN(numB)
+              ) {
+                return numA - numB
+              }
+              return (valA ?? "").localeCompare(valB ?? "", undefined, {
+                numeric: true,
+              })
+            },
             enableColumnFilter: filterChoices !== undefined,
             filterFn: multiValueFilter,
           }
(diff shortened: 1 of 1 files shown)


MiracleInvoker:
Thank you for the review!
I have verified the logic against those cases, and it will not raise any JavaScript exceptions.

In JavaScript / TypeScript, Number("string") returns NaN rather than throwing an error.
My sortingFn explicitly checks for this using !Number.isNaN(...).

If the values are strings (like "abc" in your categorical example or "value" in user attributes):

1. The conversion to Number results in NaN.
2. The if condition fails because of the isNaN check.
3. The function falls back to localeCompare, which safely handles standard string sorting.

So, mixed types or pure strings will simply fall through to the string sorting behavior, which preserves the existing functionality for non-numeric values.

image

Additionally, I noticed that `TrialTable.tsx` sets `enableSorting: false` for columns with `CategoricalDistribution`.
This means the sorting logic is effectively bypassed for categorical parameters in the UI, further ensuring no regressions for those types.

PR #1274, opened 2026-05-07
Migrate package manager from npm to pnpm

porink0424:

## Reference Issues/PRs

N/A

This PR migrates the package manager from npm to pnpm. `jupyterlab/` is excluded since it is managed by yarn (jlpm).

- Adopted `pnpm@10.33.2` because pnpm v11 was only just released and is still too new.
- pnpm does not hoist the way npm does, and a `peerDependencies` declaration alone leaves a package unable to resolve those entries in `node_modules` during its own development, build, and type-checking. To work around this, the required packages are also listed under `dependencies` / `devDependencies`.
- To keep React / MUI / Emotion as singletons, webpack's `resolve.alias` is used to point `optuna_dashboard/` and `standalone_app/` at `tslib/react/node_modules/*` for dedupe.
- `tslib/storage/` is a pure ESM package (`"type": "module"`), and since Node's ESM resolution requires file extensions on relative imports, `.js` extensions were added. (Under npm this happened to work thanks to hoisting and TS's looser resolution.)
- Security-hardening options are configured in `pnpm-workspace.yaml`. The rationale for the few options that loosen security is as follows.
  - `trustPolicyIgnoreAfter: 525600`: do not apply the trust policy to packages published more than a year ago. Without it, installs fail on pre-provenance legacy transitive deps like `semver@6.3.1` (a 2022 release reached via `@vitejs/plugin-react` → `@babel/core`), whose missing trust evidence is an artifact of age rather than a supply-chain regression.
  - `allowBuilds`: enumerates packages (`@biomejs/biome`, `esbuild`, etc.) that are explicitly permitted to run build scripts as exceptions to `strictDepBuilds`.

Follow-up tasks:

- To minimize the diff against the current repo layout, this PR keeps the existing form of running `pnpm install` separately in each subpackage. Ideally everything should be consolidated into a single pnpm workspace, which would also obviate the webpack-alias-based singleton workaround above.
- To keep the migration scope narrow, bumping deprecated packages was deferred. For example, `@mui/lab` in `tslib/react` is deprecated and left at an old version; these should also be brought up to the latest.


c-bata (approved):
LGTM!

PR #1330, opened 2026-08-30
Improve error handling in the hyperparameter importance WASM module

### LINKED ISSUE #628: Improve the error handling of WASM module calculating hyperparameter importance.
The [browser-only version of Optuna Dashboard](https://optuna-dashboard.readthedocs.io/en/latest/getting-started.html#browser-only-version-experimental) internally uses the WASM module exported by the following Rust code.

https://github.com/optuna/optuna-dashboard/blame/5c7ea7432f7b17a9535d70ee7173e41888323564/rustlib/src/lib.rs#L6-L26

However, as you can see, a lot of `unwrap` methods call exist. We should avoid using unwrap methods call.

greekera1000:

## Reference Issues/PRs
Fixes #628

The `unwrap` calls described in #628 were removed at some point, but the error handling around them still has gaps in both directions: degenerate input silently produces `NaN`, and real errors never reach the user.

### `NaN` importances on well-formed studies

`Fanova::quantify_importance` computes `variance / partial_size / tree.variance`, where `partial_size` is a product over the widths of all feature ranges. A parameter that took the same value in every trial spans a zero-width range, so that product is zero and **every** parameter's importance becomes `NaN` — not just the constant one. A constant objective zeroes `tree.variance` with the same result.

Running the current code against fanova 0.2.0:

one constant param -> [NaN, NaN]
constant objective -> [NaN]
single trial -> [NaN, NaN]

None of these are error cases; they are ordinary states for a study early in its run. They now yield zeros, with constant columns dropped before fitting so they cannot poison the rest.

### Unhandled errors in the caller

`standalone_app/src/components/StudyDetail.tsx` called `wasm_fanova_calculate` under a `// TODO: handle errors thrown by wasm_fanova_calculate` with no `catch`. Any throw — including a failed `init()` — became an unhandled promise rejection, leaving the importance panel empty with the reason visible only in the console. Errors now go through the existing `reportError` snackbar.

### Changes

- New `rustlib/src/importance.rs` holds the calculation as plain Rust with no wasm glue, making it unit-testable. `lib.rs` is now just the `#[wasm_bindgen]` wrapper.
- A typed `ImportanceError` names the offending index: `features[2][7] is not a finite number` rather than `features must be of type number[][]`. It implements `std::error::Error`, so `?` converts it to `JsError` via wasm-bindgen's blanket impl.
- Ragged columns, target-count mismatches, and non-finite values are rejected up front; `<2` trials, a constant objective, and constant parameters return zeros. A final `is_finite` check keeps any stray `NaN` out of the chart.
- `filterFunc` now uses `Number.isFinite`. This is slightly beyond the scope of the issue: the previous `!== Infinity && !== -Infinity` pair let `NaN` objective values through to the WASM module.
- The unchecked `as Optuna.TrialParam` cast is replaced with an explicit error, and the effect ignores results from a stale `study`.
- 10 unit tests covering each case above, plus `.github/workflows/rust-tests.yml`, since CI currently never runs `cargo test`.

The new workflow is the most opinionated part of this PR and easy to drop if you would rather fold `cargo test` into `typescript-tests.yml`. Happy to split the `filterFunc` change out as well if you would prefer this PR stay strictly within #628.

### DIFF
--- a/.github/workflows/rust-tests.yml
+++ b/.github/workflows/rust-tests.yml
@@ -1,0 +1,22 @@
+name: rust-checks
+on:
+  pull_request:
+    paths:
+      - '.github/workflows/rust-tests.yml'
+      - 'rustlib/**'
+
+permissions:
+  contents: read
+
+jobs:
+  test:
+    name: Unit tests for rustlib
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out code
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+
+      - name: Run tests
+        working-directory: rustlib
+        run: cargo test
--- a/rustlib/Cargo.toml
+++ b/rustlib/Cargo.toml
@@ -11,4 +11,4 @@ js-sys = "0.3.77"
 serde-wasm-bindgen = "0.6"
 
 [lib]
-crate-type = ["cdylib"]
+crate-type = ["cdylib", "rlib"]
--- a/rustlib/src/importance.rs
+++ b/rustlib/src/importance.rs
@@ -1,0 +1,288 @@
+use fanova::{FanovaOptions, RandomForestOptions};
+use std::error::Error;
+use std::fmt;
+
+/// fANOVA needs at least two observations before the objective can have any
+/// variance to attribute to the parameters.
+const MIN_TRIALS: usize = 2;
+
+/// Reason why hyperparameter importance could not be calculated.
+///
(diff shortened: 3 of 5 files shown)


c-bata:
Thanks for the contribution.

The original issue (#628) was about removing the `unwrap` calls from `rustlib/src/lib.rs`, and those calls have already been removed from the main branch. The changes in this PR go beyond the scope of #628 and would be too large to review. I’m going to close this PR.

